PR #7, opened 2019-12-13
adding spanish backend and homepage

amilich:


amilich:
Note that `settings.py` has to be returned to deploy mode before deploying

amilich:
will remove the DS_store files!
also waiting for professional translation

PR #8, opened 2019-12-15
Admin approval page

amilich:


amilich:
This page lets admins approve entries for specific states. Each user (from the /admin page) can be added to groups, like "NJ" and "VA" that lets them approve maps that have a centroid in a particular state. It uses a reverse geocoding library.

amilich:
Screen Shot 2019-12-15 at 6 22 02 PM

amilich:
^ We probably need to redo the buttons - like "Approve" or "Delete" and just not have a "Approve" button if it's already approved. We could also - for admins - remove the delete button unless they created the map themselves.

amilich:
Screen Shot 2019-12-15 at 7 24 26 PM
Adding different colors on review page.

**TODO:** We need to make sure that using `get_context_data` is completely safe in the `review` function in `views.py`.

PR #20, opened 2019-12-20
Map style

lauren-johnston:
Changes to entry map: fullscreen drawing mode, dotted lines, census blocks only appear at minzoom for aesthetic purposes

### DIFF
--- a/main/static/main/css/style.css
+++ b/main/static/main/css/style.css
@@ -227,7 +227,7 @@ color scheme. See readme.md*/
   height: 100%;
 }
 
-.draw_polygon_map {
+.draw-polygon-map {
   top: 0;
   bottom: 0;
   /*  Required by Mapbox: Must Set Absolute Height in px */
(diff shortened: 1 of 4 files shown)


amilich:
Zoom feature is great but I think census blocks should be much more visible - this is feedback from Texas where they didn't want to see roads and other stuff and really see data layers

Any way we could add toggles on/off for blocks/other data layers to the drawing page?

lauren-johnston:
Thanks - My sense is that roads and landmarks are pretty important though for people to understand the community they've drawn. It's also hard to make the census block lines much thicker without reducing visibility. I will see if I can make the names more visible though. As for the question about toggling street layers, I haven't looked into it/seen this done much before. We could toggle between street and satellite style though. Or create our own style: https://docs.mapbox.com/help/tutorials/create-a-custom-style/

amilich:
Got it - good points. I guess let's not focus on that for now?

As for the census lines, I'd try to make them a bit more visible/the dots less transparent because they are kinda hard to see or use - imagine a 70 year old person who runs the League of Women voters trying to use it :)

lauren-johnston:
sounds good will thicken the lines!

PR #21, opened 2020-01-29
Map permalinks!

amilich:


amilich:
List of changes
- Added /submission?map_id=UUID page to allow viewing a single map
- Added permalink to map to the "thanks" page
- Added links to the individual maps in the /map page

Open questions:
- Do we not want to use __startswith and the uuid to link to maps?
- How can we better restructure code to avoid copy/paste?

PR #45, opened 2020-05-06
Working whitelist, improved member form 

### LINKED ISSUE #41: Whitelist uploads not being added to users
Need to add logic when users add submissions to check if they are part of the org they are submitting to (by comparing with the whitelist)

lauren-johnston:
Whitelist fixes:

- When user uploads submission with an org selected, checks if they are a member of that org. If so, sets the admin_approved field of the entry to true. 

- If they are not a member, checks if they are on the whitelist. If so, adds them as a member and deletes whitelist entry.

Org member form fixes:

- Stops creating duplicate members when on the member creation form (local link to creation form: http://127.0.0.1:8000/dashboard/partners/[org_slug-id]/members/create)

Resolves #41

### DIFF
--- a/main/forms.py
+++ b/main/forms.py
@@ -31,6 +31,7 @@ from .models import (
     Organization,
     Campaign,
     Membership,
+    User,
 )
 from .choices import (
     RACE_CHOICES,
@@ -201,11 +202,20 @@ class CampaignForm(ModelForm):
 
 
 class MemberForm(ModelForm):
+    def __init__(self, *args, **kwargs):
+        super(MemberForm, self).__init__(*args, **kwargs)
+        self.fields["member"].widget.attrs.update({"class": "form-control"})
+        self.fields["is_org_moderator"].widget.attrs.update(
+            {"class": "form-control"}
+        )
+        self.fields["is_org_admin"].widget.attrs.update(
+            {"class": "form-control"}
+        )
+
     class Meta:
         model = Membership
         fields = [
             "member",
             "is_org_admin",
             "is_org_moderator",
-            "is_whitelisted",
         ]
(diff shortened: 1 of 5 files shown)


theodormarcu:
I checked the user flow and it works great. Approving was smooth and auto approving also worked.

I don't think we should delete users from the whitelist though. I think the whitelist can also function in the future as a way to "contact" all users directly. Updating the whitelist may confuse users because they uploaded something with X members and next time they check there are only Y members.

lauren-johnston:
To your question about the whitelist - when an email is deleted from the whitelist, the user is added to the membership table. The whitelist is not referenced anywhere (users cannot see their whitelist) but they can see their current members. I chose to delete from whitelist in order to reduce the size of the db so it doesn't duplicate small something like an email, esp because whitelists can be large. If we choose to show the whitelist ever, we can show it to them with their current member list. 

As for your point about contacting the whitelisted individuals, the whitelist they uploaded is intended to be already a mail list that they already have access to. I don't think they'll be contacting/emailing people through Representable.

theodormarcu:
The use case I have in mind is that an admin/moderator can send an email blast from representable similar to github issues/causes and users can get an official looking email with a wording along the lines of:

"YOU've been INVITED to serve your community/fight gerrymandering for organization ABC etc. Go at link XYZ to add your map"

In the vein of campaigns, I think this will be really useful in the future for removing the friction associated with using representable. I think the whitelist will become very useful and will increase in visibility then. This is why we should not remove users yet.

I don't think storage should be a concern now. The value we get from having people's emails on representable far outstrips what we'd pay for storage.

lauren-johnston:
We can still send that email when the whitelist is uploaded though? Entry doesn't get deleted until a user with that email submits an entry. And also if you wanted to send an email to everyone after we could just grab member emails + whitelist emails together. Because if a user is a member of an org (and were formerly whitelisted) it means they have submitted something. Thus we would not be sending them "YOU've been INVITED to serve your community/fight gerrymandering for organization ABC etc. Go at link XYZ to add your map"

PR #46, opened 2020-05-06
Testing suite updated

lauren-johnston:
Committing basic testing suite


theodormarcu:
Can you add more documentation about what this does?

lauren-johnston:
@theodormarcu I am trying to get Github CI working! Issue #54 describes part of the problem here. The first issue is that geo libraries needed to be excluded because they couldn't be downloaded using requirements.txt. The second problem is that the testing suite cannot find the environment variables. I think we want dummy variables for the env variables so as not to have the actual prod variables in testing. Haven't figured out the best way to include them yet though.

theodormarcu:
Oh that makes complete sense! Do you think Travis CI/Circle CI might be more powerful in that case? I know they can do brew-related stuff.

GitHub CI might be for simpler apps.

lauren-johnston:
Yeah I wanted to give Github CI a shot because it is completely integrated with github and we can see the CI results right here in our git dashboard. It also seems much more developed from the last time I tried it and has documentation here: https://help.github.com/en/actions/building-and-testing-code-with-continuous-integration/about-continuous-integration. I set up github CI through the Actions tab - immediately there was an option to configure CI for a Django app which was a good sign. If we poke around and it doesn't seem like the right solution for us down to go with another CI!

Issue #42, opened 2020-05-06
Entry page - Census blocks inside and outside of community same color 


lauren-johnston:
User feedback: "What is the lighter blue area to the west of my polygon and how can I get rid of it?" Replicated issue below with screenshot.

<img width="885" alt="Screen Shot 2020-05-05 at 10 16 56 PM" src="https://user-images.githubusercontent.com/8892509/81140473-50772d80-8f1e-11ea-93b8-bc8bbcb3a99e.png">

k-y-le:
![image](https://user-images.githubusercontent.com/41943646/81167106-f76bd000-8fd7-11ea-9173-39068475baec.png)
they may have encountered something like this (unrelated to census blocks)

PR #47, opened 2020-05-06
a collection of small changes to visuals

k-y-le:
- updated wording on entry page (not just zipcode)

- made default map of the entire US (entry page and partner map page)

- changed census block shading to be gray + more transparent, generally (TODO: update to a log scale, maybe?)

- updated partner map page to display org name, in place of spacebar

- partner maps zoom to where communities have been drawn (if no approved communities, remains on map of whole us)

- updated community visualizations to be more distinguishable

- updated census blocks to be less obstructive

- census block and legislature layers now have a minzoom, so they don't clutter the map when zooming out (also reduces lag) (TODO: explain this in tooltip)

image
image
image

### DIFF
--- a/main/static/main/js/geo.js
+++ b/main/static/main/js/geo.js
@@ -240,8 +240,8 @@ document.addEventListener(
 var map = new mapboxgl.Map({
   container: "map", // container id
   style: "mapbox://styles/mapbox/streets-v11", //hosted style id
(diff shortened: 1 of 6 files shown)


lauren-johnston:
Looks great! These style changes on the maps look really good.

lauren-johnston:
Merging this ahead of meeting today! Also: this PR resolves issue #42

PR #75, opened 2020-05-09
Sdev

arorasomya:
Added a separate view for the thanks page when the user selects a particular org. Changed the link on the thanks page from the map uri to the partner org's map page. 
Also changed the link of the thanks page, something like - http://representable.org/somyas-liberty-org/thanks.

image

image


lauren-johnston:
Census block snapping now working! One thing I realized is that if people are not approved to submit to an org their submission will have to go through a review. So it may be better to: 1) keep the original submission link (in case they want to see the submission immediately) and 2) check if user is already a member of the org and if not say "your submission has been sent to <org name> for review and if it is approved will soon be available on the public map page <org map link>!

lauren-johnston:
I added a quick disclaimer that map will be available once changes are approved!
Screen Shot 2020-05-20 at 3 29 59 PM

lauren-johnston:
Approving knowing that this may change slightly with campaigns however we will only need to make small modifications to this code to make that happen.

PR #76, opened 2020-05-09
cleanuo of geo.js - commented out censuslines code etc

arorasomya:


lauren-johnston:
Sweet! Does this fix the Mapbox errors? Also we may want to remove this code instead of commenting it in production for clarity unless we think we'll use it again later.

arorasomya:
I did remove the commented out code. It was the code when we moved from census lines to census block shading. but the method was still there. 
re: mapbox issues: no that's still a problem. i can look more into them later

lauren-johnston:
Oh gotcha yeah I misread the title but I see the code is removed now!

PR #77, opened 2020-05-09
bug for polygon too big for michigan - updated area for all states

arorasomya:
Bug mentioned by Sandy - community too big to draw - because we were using the area for NJ.
Expanded this check for all 50 states.

image

Notice the print statement to the right that pushes the error for the michigan coi and deletes a vertex to take care of it.


arorasomya:
Yah i think it's because of the mapbox issue that i told you about earlier

PR #93, opened 2020-05-19
Campaign model changes, basic dashboard views

lauren-johnston:
Updates:
- Changed campaign model to the following fields (see spec doc):
    - id: uuid for campaigns
    - name: name of the campaign
    - state: the state of the campaign
    - description: description of the campaign
    - organization: organization hosting the campaign
    - is_active: is the campaign active
- Changed the id field for campaigns to uuid for future security needs/creating unique links
- Modified basic views:
    - create campaign 
    - update campaign details
    - add campaigns to dash home 
- Still to do:
    - create custom campaign/entry page link
    - create campaign map link
    - add campaign field to org map review
    - create public campaign pages (out of scope for this push)
    - frontend for campaign form

Screen Shot 2020-05-19 at 11 07 30 AM
Screen Shot 2020-05-19 at 11 07 42 AM
Screen Shot 2020-05-19 at 11 08 04 AM

Resolves #92

PR #94, opened 2020-05-19
Adds custom campaign links

### LINKED ISSUE #87: Hide Campaign field in entry page when users are using a preset campaign link
The campaign select field is prefilled and hidden from users who are using a special campaign specific link to add entries

lauren-johnston:
Note: this PR is related to #93 .

- Creates a custom unique token for each campaign in a new "CampaignToken" model
- Displays the custom entry page + token link on campaign dashboard
- When entry page has token string, hides the organization field
- On entry submission, when token is included retrieves token campaign and organization and updates fields 
- On entry submission, when token is included AUTO approves submission.

Main steps to replicate this:
- Create org (or use existing org)
- Create campaign
- Follow entry page link
- Try creating an entry
- Check out postico to see that this entry has the correct campaign + org associated
- Check out org review dashboard to see that entry has been auto-approved

Also worth verifying:
- Entry page still works without token specified

Still to do:
- Add campaign map
- Add campaign field to review 
- Add campaign link on state page 

Screen Shot 2020-05-19 at 1 15 25 PM

Resolves #87


indraneel:
why do campaign tokens need to be UUIDs? I'm thinking that, from the perspective of somebody sending around/opening a URL, having it be a long UUID string might be unwieldy.

I think I understand the auth constraints (if token exists: approve) but it also looks like the token is not unique to a user, but rather it is unique to the campaign? in which case I wonder if a simpler or more human readable token might do the job?

lauren-johnston:
Thanks for taking a look @indraneel! Campaign tokens are generated with this function in utils.py: 

```python
def generate_unique_token(campaign):
    salt = uuid.uuid4().hex + campaign
    return hashlib.sha256(salt.encode("utf-8")).hexdigest()
```
My logic for making the token long is so it is not guessable. It's also coded this way so that we can easily turn it into a one-time link that also takes the user's email as salt in the future. (You're right, it currently does not take the user's email to create a token). That way, we would send unique links to the emails of people participating in the campaign and only allow users with that email to use the token. I didn't think the length of the token would be a problem given that it's only used to pre-fill the entry page (and not for map links) and also Preeti will be directly working with our partners to share the link directly with them.

indraneel:
got it! that was the context i was missing, thanks!

lauren-johnston:
This does bring up an interesting point though that not every campaign will be private/closed to an organizations members and lists and in those cases people will want simple links to share rather than longer links

lauren-johnston:
Update: I just added a small commit that adds org permissions mixins for campaign views. This means that only org admins will be able to update/create campaigns. Sorry for any inconvenience this causes; I considered creating a separate PR but realized it should really be in this one.

PR #97, opened 2020-05-20
Entry form redesign backend

arorasomya:
Changes in this PR
- new fields added to sync up with the paper and pencil form in the backend
- address table created to store 1 address per entry

image

image

image

image


theodormarcu:
Can you first check why Travis CI fails?

lauren-johnston:
> Can you first check why Travis CI fails?

When I click on the TravisCI build I see `ModuleNotFoundError: No module named 'phonenumber_field'` @arorasomya

arorasomya:
> > Can you first check why Travis CI fails?
> 
> When I click on the TravisCI build I see `ModuleNotFoundError: No module named 'phonenumber_field'` @arorasomya

@lauren-johnston  I added a new module in requirements.txt. Maybe it wasn't downlaoded/installed in the travis ci build environment?

arorasomya:
> > > Can you first check why Travis CI fails?
> > 
> > 
> > When I click on the TravisCI build I see `ModuleNotFoundError: No module named 'phonenumber_field'` @arorasomya
> 
> @lauren-johnston I added a new module in requirements.txt. Maybe it wasn't downlaoded/installed in the travis ci build environment?

Oh I found it. I changed the type of phone number field I was using and it's catching the old one in one of the mirgations.

arorasomya:
@lauren-johnston @theodormarcu resolved the check fail issue. Unsure why it's only showing one check and saying "all checks have passed"

theodormarcu:
Everything looks great except for the phone number. There is no DB validation for the kind of data that should be in there. I think that we need that. In addition to that, we need some sort of rudimentary form validation to let the user know that they can only write a phone number there.

Screenshots shows how I've been able to save a phone number and how the form does not complain about it.

Screen Shot 2020-05-21 at 00 18 00
Screen Shot 2020-05-21 at 00 20 12

lauren-johnston:
Lots of ideas here for phone number validation: Seems like the easiest is regex (esp since we only need to handle US phone numbers). https://stackoverflow.com/questions/19130942/whats-the-best-way-to-store-phone-number-in-django-models. Also - @theodormarcu is this something that could be added in a subsequent PR?

lauren-johnston:
Actually nvm I agree this should probably be resolved in this PR - I'm also not sure if importing a phone number model is necessary when we're only handling US phone numbers. One thing we could do is remove this phone number field entirely in the interest of making sure Kyle can start working on adding additional fields to the review page.

arorasomya:
> Actually nvm I agree this should probably be resolved in this PR - I'm also not sure if importing a phone number model is necessary when we're only handling US phone numbers. One thing we could do is remove this phone number field entirely in the interest of making sure Kyle can start working on adding additional fields to the review page.

I just added a validator that reloads the page if the phone number is not in the correct format (no letters allowed) - it's a specific kind of format - E164

arorasomya:
Oh and also removed the separate extension field

lauren-johnston:
> > Actually nvm I agree this should probably be resolved in this PR - I'm also not sure if importing a phone number model is necessary when we're only handling US phone numbers. One thing we could do is remove this phone number field entirely in the interest of making sure Kyle can start working on adding additional fields to the review page.
> 
> I just added a validator that reloads the page if the phone number is not in the correct format (no letters allowed) - it's a specific kind of format - E164

Ok sounds good to me, thanks so much for making these changes! I took a look at the file changes and they looked good but did not run it. Let me know if you want/need an additional reviewer and I'm happy to do an actual review.

arorasomya:
> > > Actually nvm I agree this should probably be resolved in this PR - I'm also not sure if importing a phone number model is necessary when we're only handling US phone numbers. One thing we could do is remove this phone number field entirely in the interest of making sure Kyle can start working on adding additional fields to the review page.
> > 
> > 
> > I just added a validator that reloads the page if the phone number is not in the correct format (no letters allowed) - it's a specific kind of format - E164
> 
> Ok sounds good to me, thanks so much for making these changes! I took a look at the file changes and they looked good but did not run it. Let me know if you want/need an additional reviewer and I'm happy to do an actual review.

That would be great! It wouldn't hurt to have another pair of eyes :)

arorasomya:
> Everything looks great except for the phone number. There is no DB validation for the kind of data that should be in there. I think that we need that. In addition to that, we need some sort of rudimentary form validation to let the user know that they can only write a phone number there.
> 
> Screenshots show how I've been able to save a phone number and how the form does not complain about it.
> 
> Screen Shot 2020-05-21 at 00 18 00
> 
> Screen Shot 2020-05-21 at 00 20 12
By "form validation to let the user know that they can only write a phone number there." did you mean add some kind of a style change to the field?

theodormarcu:
I meant something to say that you can only write a phone number, but I now realize that it's more of a front-end thing and it can be implemented in a subsequent PR! Will Approve it :) 

Next steps: I merge my PR and then we do this one.

PR #102, opened 2020-05-21
new user flows - map select + state pages

k-y-le:
NOTE: this is not totally complete - still need content for michigan page
- added state map svg (jquery) with highlighted states on main page. select states via dropdown of hover. responsive for mobile too.

- added "partner with us" portion of landing page, linking to partner welcome page + new graphic

- states: all state links currently lead to the default entry page, except michigan

- michigan page: hardcoded a michigan page in urls -- page is currently a placeholder as we wait for content. will in the future have specific info abt redistricting in michigan, and a form to check any public campaigns in the state before heading to the entry page. for this rollout we are not including the campaigns and instead linking to just the entry page. future steps also include implementing templates such that we don't need to hardcode an html file for every state

- included a library of .svg files for each state - for use on state pages or anywhere else

- zoom on all map + review pages now goes closer to the community

- cleaned up map page a bit: removed searchbar, added beta disclosure, and implemented zoom functionality

image
image
image
image


lauren-johnston:
Also there was a merge conflict with .DS_Store so I added .DS_Store to the .gitignore and removed all the .DS_Store files.  Now you should be able to merge without conflicts :)

PR #107, opened 2020-05-22
Campaign page

lauren-johnston:
**Key changes:**
- Adds unique slug to campaign model (in same format as org slugs)
- Creates campaign page with slug 
- Adds "General information about MI" if state == MI on campaign page
- Prefills entry when campaign slug supplied (does not autoapprove)
- Checks if user is on campaign's organization's whitelist or if user and if so approves submission
- Temporarily stops entry page tokens from autoapproving submissions until we know exactly how tokens will be used 

**Key steps to replicate:**

- Run migrations
- Make a campaign from an organization 
- Navigate to the campaign page (linked in "Gather submissions")
- Try creating an entry from here
- Verify that organization dropdown does not appear using custom entry link
- Verify in Postico that entry has the right campaign & org

**To Do:** 
- Add real MI content on campaign page instead of placeholder.
- Work with @k-y-le to create a campaign map page. Currently, on the campaign page only the org is linked.

**Screenshots:**
Screen Shot 2020-05-21 at 9 45 06 PM
Screen Shot 2020-05-21 at 9 35 16 PM


lauren-johnston:
> Initial Review Only: Website crashed when making new campaign.
> 
Looks like a migrations issue with creating a new unique field! Taking a look. Hopefully the recent commits address this.

Issue #73, opened 2020-05-08
Add additional community identifying questions


lauren-johnston:
**Is your feature request related to a problem? Please describe.**
Users need to be able to add more specific information about their community's interests. 

**Describe the solution you'd like**
I think this request can be split into two parts: A) they want some kind of community-identifying questions before asking for a map entry B) they want their specific questions to be those community identifying questions
At minimum, I think they make a good point about part (A) — it might help people to think about their community first in order to think about the geography of that community. Part (B) might have some value too but I think we can figure that out later
-- @indraneel 

**Describe alternatives you've considered**
Discussed using tags, other issue categories. Ultimately, it seems like free form text is the best way to gather COI submissions.

**Additional context**
![81345458-859e8f80-9086-11ea-8fb0-5fc722daf58f](https://user-images.githubusercontent.com/8892509/81440675-4b37ff80-9125-11ea-8a06-eeaea312cde6.png)

![image](https://user-images.githubusercontent.com/8892509/81440781-715d9f80-9125-11ea-8a83-f3ab122cb9f6.png)

Issue #59, opened 2020-05-06
Entry Page/Website: Clarify That Submitting Entry Does NOT Submit to Districting Commission


theodormarcu:
- We need to include a blurb at the end to encourage people to download/print their maps.
- Clarify that either them or the organizations need to submit the info to the commission.

psiyer7:
**Other Clarifications that are needed Running List (@ various points):** 
2. When drawing -- your community might not look exactly how you drew it, it will snap to census blocks
3. When submitting -- Representable reserves the right to remove data based on moderation

theodormarcu:
@k-y-le Can I take this? I might include some changes as well as part of new front end entry.

k-y-le:
@theodormarcu yes feel free!

theodormarcu:
Closing issue per PR.

PR #108, opened 2020-05-22
Re-organized Entry Form Sections

### LINKED ISSUE #104: Searching for location on entry page should automatically go to next step
**Describe the bug**
When searching for a starting location on the entry page, there isn't any signifier that the search went through without pressing next.

**To Reproduce**
1. Search for community (press enter)
2. nothing happens lol

**Expected behavior**
automatically go to map page

theodormarcu:
Resolves #104, #59, #73.

**Key Changes:**
- Entry form sections follow the pencil and paper form more closely (#73).
- Adds disclaimer that form is not submitted to commission and instructions (#59).
- Combines map and geocoder in one section to create an easier transition (#104).
- Moves map last so that users can build context before drawing.
- Updates map fields to make them more accessible.
- Added personal information disclaimer.

**How to test:**
1. Go to entry form.
2. Create an entry. Make sure JS works for section traversal ("Next", "Back", "Hide/Edit Section").
3. Fill form. 
4. Read text.
5. Check that map only toggles after you search for a location in the geocoder.

**To Do/Next Steps:**
1. Discuss content with PGP/VNP. Is our disclaimer clear enough?
2. Is our personal information disclaimer good enough?

**Screenshots:**

Screen Shot 2020-05-22 at 01 17 10
Screen Shot 2020-05-22 at 01 17 03
Screen Shot 2020-05-22 at 01 18 41


theodormarcu:
Fixed the issues you mentioned. Now, if there's an error with the form, it will be displayed to the user if they try to submit it again. This replicates prior behavior.

Note that the form validation is still a bit janky. I will submit an issue that details what should happen. That will be a separate PR because it affects other things as well.

theodormarcu:
That's a good catch. I'll add it to the issue I mentioned. Thank you!

arorasomya:
Checked out and reviewed as well! 
@lauren-johnston to clarify the issue you flagged: the community that you drew is actually not bigger than half the state area of NJ, which is why it passes. That's the threshold we are checking against at the moment. I tried doing the same and once you drag the corners to make the community bigger and it surpasses that threshold, one of the vertices is automatically deleted for error handling purposes. Maybe this is an issue of changing that threshold?

theodormarcu:
Oh thanks for checking that Somya! I thought the handlers were being called wrongly, but I'm glad to hear it's working as intended.

arorasomya:
Also the other memory issue you mentioned, is something that I already made an issue for and it is on the backend - mapbox side. We really need to fix it for michigan since the communities will on average be on the larger side.

PR #109, opened 2020-05-22
Campaign map/review page

lauren-johnston:
**Main changes:**
- Adds url for campaign map (parameter to existing org map)
- Adds url for campaign review (parameter to existing org review)
- Campaign dashboard home updated

**Key steps to replicate:**
- Create campaign
- Go to review page
- Check out links to campaign page, review page, and map
- Create a submission with this campaign
- See that campaign map now updated with your community

**Screenshots:**
Screen Shot 2020-05-22 at 12 46 45 AM
Screen Shot 2020-05-22 at 12 23 42 AM

### DIFF
--- a/main/templates/main/campaigns/page.html
+++ b/main/templates/main/campaigns/page.html
@@ -16,7 +16,7 @@
       {% endif %}
       <a class="btn btn-primary my-3" href="{% url 'main:entry' campaign=object.slug %}" role="button">Draw my community</a>
 
-      <a class="btn btn-outline-primary my-3" href="{% url 'main:partner_map' object.organization.slug %}" role="button">View map of communities</a>
+      <a class="btn btn-outline-primary my-3" href="{% url 'main:partner_map' object.organization.slug object.slug %}" role="button">View map of communities</a>
     </div>
     </div>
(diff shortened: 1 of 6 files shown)


lauren-johnston:
> The PR worked as expected. Approving. :)
> 
> Two clarification questions (that can be resolved in a future PR if necessary):
> 
> 1. Do we want to allow people to add submissions outside of the campaign's home state?

Good point! We talked about this being out of scope for the current push but our partner definitely wanted a clear way to make sure MI communities are the ones being submitted.

I think we discussed that we would allow submissions from everywhere but only auto-approve those that are both whitelisted AND within the state. We would clearly indication on the review page communities that were in state vs. out of state (potentially pushing out of state communities to the bottom).

> 2. Do we want to continue allowing people to submit to organization from the normal entry page?
Was also thinking about this! I think we want to encourage a flow where a campaign is created and the pre-filled entry link is shared instead of submitting directly to the org. And the org map page would be organized by each campaign with each campaign in the org being toggle-able. This merits further discussion being this PR though I will try to simplify the user flow from creating org -> creating campaign -> share campaign.

PR #110, opened 2020-05-22
submission page download geojson

k-y-le:
- now you can download your community as a geojson file ! on the submission page there is a button under the community name + reason fields (soon to be updated for other stuff as well). pressing this button prompts a download of community_ID.geojson! hooray! (this is viewable like any other JSON: prettyprint or load it into qGIS :--)

- added export views + urls that handles downloading the community -- this should work for any community as long as we have the entry_id for the url

image
image
beautiful midland michigan community in qGIS 🦙 (the lines in the middle are bc of inconsistencies in census block data and unavoidable)


lauren-johnston:
Also I did verify the GeoJson file downloaded is correct (phone edited out for privacy) on http://geojson.io/! Screen Shot 2020-05-22 at 11 19 56 AM

theodormarcu:
I agree with Lauren. GeoJSON should only contain the census blocks map and the user drawn map, but not PI (personal information). Since JSON's are essentially dicts, you should be able to strip that out. Probably requires minor change to main.py for the ReviewView, so that Django does not send all the info from the db to the context variable that gets parsed by the Django template tags.

In the future we can allow users to download GeoJSONs with all the info if done from the my entries page or a campaign management page.

k-y-le:
I think the GeoJSON should include the community name + responses to the community reasons + interests parts of the forms? assuming that that's public on the map page, I think it makes sense to include that. I agree that phone number + name should be left out

Issue #111, opened 2020-05-22
Better Form Validation


theodormarcu:
The form validation system is janky. If the user makes a mistake with the form, there are some bits of inconsistent behavior:
- If they forgets to fill a field, that field will be validated by the browser and it will force to user to take a look (desirable).
- If the user makes a mistake with the map, a custom message will appear (desirable), but if they try to submit it, all sections collapse (undesirable).

**Describe the solution you'd like**
Form validation should be redone from the ground-up, according to the new form. This includes showing a list of errors at the beginning of the page (on page refresh), expanding sections with errors, describing errors, etc.

First, we should have some custom validation for these things on form submit. This is client-side validation, done through a JS listener. Additionally, we should also ensure that any validation that should happen in forms.py happens there. Currently, the only possible validation is for kinks in the polygon, which works great. Howver, we may want to add more descriptive messages for map errors, address fields, etc.

theodormarcu:
Adding other relevant issues: https://github.com/Representable/representable/pull/108#issuecomment-632765940

We need clearer thresholds and more descriptive error messages for when the community is too big.

theodormarcu:
This is also relevant to #103

theodormarcu:
If drawn map is good and there's a problem with another field, then the map will tell you that it looks great but will not give you the chance to review it again/edit it.
<img width="904" alt="Screen Shot 2020-05-23 at 11 00 04" src="https://user-images.githubusercontent.com/8714338/82733812-92f18600-9ce4-11ea-9b5c-61ae1268b4a8.png">

Issue #105, opened 2020-05-22
Entry Page on Failed Submit Does Not Highlight Census Blocks


theodormarcu:
**Describe the bug**
If the form submission fails for any reason, the map does not highlight the right census blocks.

**To Reproduce**
1. Fill entry form.
2. Write text for phone number (example).
3. Submit.
4. Check map. Map will only show user poly, but not census blocks.

**Expected behavior**
Map should show both user poly and census blocks.

**Screenshots**
<img width="925" alt="Screen Shot 2020-05-21 at 23 15 52" src="https://user-images.githubusercontent.com/8714338/82627721-05654780-9bb9-11ea-919d-472541e73439.png">

PR #115, opened 2020-05-23
Entry Page - Phone number validation - frontend and backend

arorasomya:
- Added phone number validation
- clean function added in the backend for the phone number field
- alert added on the form

image

image

### DIFF
--- a/main/forms.py
+++ b/main/forms.py
@@ -175,6 +175,11 @@ class CommunityForm(ModelForm):
         #     raise forms.ValidationError("Area is too big.", code="area_too_big")
         return data
 
+    def clean_user_phone(self, *args, **kwargs):
+        phone = self.cleaned_data.get("user_phone")
+        if not phone.is_usa:
+            raise forms.ValidationError("Invalid phone number.", code="invalid_phone_number")
+        return phone
 
 class DeletionForm(ModelForm):
     class Meta:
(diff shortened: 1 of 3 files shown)


lauren-johnston:
I just checked this out locally and not seeing an error message when I tried to validate 1212131311222

arorasomya:
> 1212131311222

image
i copied and pasted your number and it showed up on mine. Let me look into this.

theodormarcu:
Currently reviewing this.

arorasomya:
> It worked for me.
> 
> A few considerations:
> 
> * I think there's an argument for using the full-featured phone number field. For example, here you could simply add a phone number that goes 123-156-7890, which is wrong per the standard since the 156 part should never start that way (phone numbers also have to follow a certain number of considerations). I don' think this is necessary now, but it's something to keep in mind for the future.
> * Alerts should go under the field title (e.g. the Phone is invalid alert should go under the phone number title for the field). This can be resolved with #105 and #111

- that's a valid point -> although I'm not sure if the full featured one necessarily keeps track of the area codes. From what it looked like, it added functionality for International numbers. But I can look further into this and resolve in a subsequent pr!

theodormarcu:
Great point! Definitely not a great priority rn.

*Theodor Marcu* | Princeton University | Computer Science | +1 (609) 933-5057 | theodormarcu.com ( http://theodormarcu.com )

PR #116, opened 2020-05-23
Interest fields validation - backend and frontend

arorasomya:
- Added validation to check if at least one of the interests fields has been filled out
- combined the error checks in one clean method in forms.py
- included user_polygon validation in the clean method as well
- included a line from the phone validation to make sure merge doesn't remove imp code
image
image

A lot of the fields are not filled out properly - interests blank, no map drawn (keep in mind phone number validation is a separate pr and is awaiting review)

image
 - the expansion of the fields needs to happen according to the issue theo created - will come in a subsequent pr


theodormarcu:
Currently reviewing this.

PR #117, opened 2020-05-23
added examples for the three interest fields

arorasomya:
- added an examples button (tooltip) that shows examples in line with the paper and pencil form
- will add it for other_considerations post consolidation of the fields in one section
image

image

image


theodormarcu:
Reviewing this right now.

theodormarcu:
There are some problems with the mobile version. You can debug them using the Chrome console and see device on mobile there.

Screen Shot 2020-05-23 at 11 05 18
Screen Shot 2020-05-23 at 11 05 07

arorasomya:
So @lauren-johnston suggested doing a expandable/collapsable component that is triggered when u click on the link for the examples. Looks like this:
image
 but without the card header aspect. 
image
Would say "show examples" instead of "edit details"

arorasomya:
Updates:
- changed show examples from a button to accordion style section expander
- added examples as placeholders in the inputs
image

PR #122, opened 2020-05-24
Form Consolidation (Combining interests/Expanding Sections/Removing Buttons)

theodormarcu:
- Removed Next/Previous Buttons.
- Removed Edit Section/Hide Section buttons.
- Expanded all sections by default.
- Consolidated Interests into one section.
- Added community name after issues field.

To test:
- Open entry page and fill form.
- Pay attention to examples and questions. Please raise any issues that may arise with 

Screen Shot 2020-05-24 at 01 54 20
Screen Shot 2020-05-24 at 01 54 30

### DIFF
--- a/main/forms.py
+++ b/main/forms.py
@@ -128,32 +128,45 @@ class CommunityForm(ModelForm):
             ),
             "entry_reason": forms.Textarea(
                 attrs={
-                    "placeholder": "This community is brought together by...",
+                    "placeholder": "Example: This community is brought together by ... and we'd like the commission to\
+                                    keep us in the same district so we can be represented by one person",
                     "rows": 5,
                 }
             ),
             "cultural_interests": forms.Textarea(
-                attrs={"placeholder": "Examples: Age groups, religious groups," \
-                    " arts and cultural centers, languages, immigration status, " \
-                    "ethnic groups such as Detroit’s Mexican Town or " \
-                    "Dearborn’s Middle Eastern community, historic settlements " \
-                    "such as Leland’s Fishtown.", "rows": 5}
+                attrs={
+                    "placeholder": "Examples: Age groups, religious groups,"
+                    " arts and cultural centers, languages, immigration status, "
+                    "ethnic groups such as Detroit’s Mexican Town or "
+                    "Dearborn’s Middle Eastern community, historic settlements "
+                    "such as Leland’s Fishtown.",
+                    "rows": 5,
+                }
             ),
             "economic_interests": forms.Textarea(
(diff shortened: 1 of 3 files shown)


theodormarcu:
Replying to comments above.

> "Recommended" in turquoise text was a bit confusing to me initially. I think it's helpful when it's a warning color like the red it was before but I realize you probably changed it to avoid the connotation of a required question, which is valid. If you want to keep it blue would suggest making it the same shade of blue as "show examples" (the primary color) otherwise I think they clash a bit being so close together.
> "Recommended" appears larger than "required" across the form. For consistency, I think it would make sense to add the "recommended" text to the question rather than section title.

Agreed. Will implement this.

> Section titles have inconsistent use of "your". I guess I would suggest simplifying and removing "your" from section titles altogether. "your community's shared interests" -> "shared interests" etc. I don't have super strong feelings on this though as long as it's consistent and concise.

I tried to strike a friendlier tone, but I agree that consistency is important.

> (Warning - this one's pedantic) Capitalization of community is inconsistent ("Community" vs "community"). Would suggest going with lower case as it is not a proper noun, though I'm not a grammar expert.
 Nah good point!

> Placeholder text in the screenshot below has strange spacing and also the text is confusing with the "...". I also wonder if we should give examples of potential issues but not prescribe a solution like "we'd like the redistricting commission to keep us in the same district". I think they need to come to this sort of conclusion independently or we could be introducing bias. Also, I wonder if there will be people filling this out who know the issues facing their community but are not sure how the commission could help in which case we should rephrase the question to be more like "what are the issues facing your community and do you have any suggestions for how the redistricting commission in your state could help address these issues?" I realize I suggested the original wording but just now when filling out the form realized this could be an issue.

I like this phrasing much more too.

theodormarcu:
>Placeholder text in the screenshot below has strange spacing and also the text is confusing with the "...". I also wonder if we should give examples of potential issues but not prescribe a solution like "we'd like the redistricting commission to keep us in the same district". I think they need to come to this sort of conclusion independently or we could be introducing bias. Also, I wonder if there will be people filling this out who know the issues facing their community but are not sure how the commission could help in which case we should rephrase the question to be more like "what are the issues facing your community and do you have any suggestions for how the redistricting commission in your state could help address these issues?" I realize I suggested the original wording but just now when filling out the form realized this could be an issue.

Also, with regard to the point of prescribing solutions, I think that's fine based on our conversations with Connie et al. This is especially because this bias is baked into your reason for coming to Representable.org. Also, after getting that list of issues from her, I realized that most of them sounded the same ("we want to be in district x or separate from people who do y"). I'm open to recommendations though.

theodormarcu:
Update: I added Connie's example.

PR #126, opened 2020-05-24
visualization updates

k-y-le:
A lot of changes ! Here are the main ones:

- removed styling from front page dropdown

- standardized visualization format across pages: maps, reviews, and submission page

- included new information (if available), will need to be updated once other model changes are pushed

- new info available via small "Show More" button on each page. what's otherwise visible is the name (only on org/campaign review page) and org/campaign (when on general map page) or campaign (when on org page)

- click on a community to zoom to it

- updated initial zoom to bounds of first community

- changed map pages to light theme (less obstruction when viewing larger numbers of communities)

- made census block lines less ugly

- made review pages more mobile responsive

- refactored most of the javascript: deleted partner_map.js (redundant), made components folder with constants, re-did a bunch of functions to be shorter/cleaner, changed views to be more concise/take out unnecessary parts

- TODO: update header css for org/campaign info, give community info part of visualizations a more distinctive look, filtering communities options, search, colors (???), change zoom for organizations/campaigns/state specific map pages, make census block lines thicker at close zooms

image
image
image
image


theodormarcu:
Currently reviewing this.

theodormarcu:
This is amazing Kyle! I'm a big fan of the visualization stuff. I have a few nitty-gritty comments.
1. The Organization and Campaign maps only display the list of entries after the map zooms in. I think the communities should load as early as possible, because I was initially very confused and I have a feeling our users will too.
2. My understanding is that that any user should see the name and address fields on "My Entries". This also applies to the Campaign/Organization map pages for Moderators/Admins. However, I could not replicate this. Is this by design/a priority for an upcoming PR?

k-y-le:
> 1. The Organization and Campaign maps only display the list of entries after the map zooms in. I think the communities should load as early as possible, because I was initially very confused and I have a feeling our users will too.
I agree it can be confusing -- I can change it so that it shows all entries before only showing the ones that are zoomed in, but I think that is also confusing. It's not possible to have the communities on the map (search while dragging) shown at first, because the page has to wait for mapbox to render the features. The initial animation will get faster as we zoom in on the state only, but I think perhaps a lil loading icon is best here?
> 2. My understanding is that that any user should see the name and address fields on "My Entries". This also applies to the Campaign/Organization map pages for Moderators/Admins. However, I could not replicate this. Is this by design/a priority for an upcoming PR?
I didn't include these details on My Entries because I assumed it would be redundant since all the communities visible will be coming from the same person-- it's easy to include though if we still think it's necessary. I defaulted to keeping personal details off all the map pages but I can make it visible to admins/mods.

side note: I was a little confused by the models and finding the address, so i didn't include it in this PR -- once updating it for the latest migrations I'll add

lauren-johnston:
> I agree it can be confusing -- I can change it so that it shows all entries before only showing the ones that are zoomed in, but I think that is also confusing. It's not possible to have the communities on the map (search while dragging) shown at first, because the page has to wait for mapbox to render the features. The initial animation will get faster as we zoom in on the state only, but I think perhaps a lil loading icon is best here?

The loading icon idea sounds good. I think 2) can be addressed in future PRs.

k-y-le:
Pushed these changes -- currently says "Loading Communities..." at first -- will update in the future so that it looks nicer with a lil animation or something

theodormarcu:
1. Really ingenious! I love it. 
2. Future PR it is! We should create an issue though. @k-y-le

PR #135, opened 2020-05-25
visualizations fully updated

k-y-le:
- Visualizations now include the other issues field, addresses (for those with permission to see them), and have been cleaned up a bit to look nicer

** to see **
go to any map page yee haw 🤠 

image
image

### DIFF
--- a/main/static/main/js/map.js
+++ b/main/static/main/js/map.js
@@ -172,7 +172,7 @@ map.on("load", function () {
     var northEast = new mapboxgl.LngLat(fit['_northEast']['lat'], fit['_northEast']['lng']);
     community_bounds[obj] = new mapboxgl.LngLatBounds(southWest, northEast)
     if (zooming) {
-      map.fitBounds(community_bounds[obj], {padding: 100});
+      map.fitBounds(community_bounds[obj], {padding: 200});
       zooming = false;
     }
     // draw the polygon
@@ -261,7 +261,7 @@ map.on("load", function () {
     map.setPaintProperty(this.id + "line", "line-width", 2);
     map.setPaintProperty(this.id, "fill-opacity", 0.15);
   });
-  $(".loading").delay(1200).fadeOut(1000);
+  $(".loading").delay(1500).fadeOut(1000);
 });
 
 // on click, zoom to community
--- a/main/templates/main/dashboard/partners/review.html
+++ b/main/templates/main/dashboard/partners/review.html
@@ -63,44 +63,56 @@
             {% for c in communities %}
             <tr class="community-review-span" id={{c.entry_ID}}>
               <td>
-                <span class="font-weight-light text-uppercase entry-name">
+                <span class="text-uppercase entry-name">
                   {% with link_text="/submission?map_id="|addstr:c.entry_ID|slice:27 %}
                   <a href={{link_text}}>{{c.entry_name}}</a>
                   {% endwith %}
                 </span>
+                <hr class="my-1">
+                {% if campaign is None %}
+                {% if c.campaign %}
+                <span class="badge badge-pill badge-secondary">{{c.campaign}}</span>
                 <br>
+                {% endif %}
+                {% endif %}
+                <span class="small text-muted">
                 {% if c.user_name %}
                 {{ c.user_name }}
                 {% else %}
                 {{ c.user.username }}
+                {% endif %} |
+                {% if c.user_phone %}
+                {{ c.user_phone }} |
+                {{streets|get_item:c.entry_ID}},
+                {{cities|get_item:c.entry_ID}}
                 {% endif %}
+                </span>
                 <br>
-                {% if campaign is None %}
-                {% if c.campaign %}
-                <b>Campaign: </b>{{c.campaign}}
-                <br>
-                {% endif %}
-                {% endif %}
-                <span class="comm-content">
+                <span class="font-weight-light comm-content">
                   <span class="more-content">
-                    {{c.user_phone}}
-                    <br>
-                    <!-- <b>Address: </b>{{c.user.address}}
-                    <br> -->
-                    {% if c.entry_reason %}
-                    <b> Entry Reason</b><br>{{c.entry_reason}}
-                    <br>
-                    {% endif %}
                     {% if c.cultural_interests %}
-                    <b> Cultural Interests</b><br>{{c.cultural_interests}}
-                    <br>
-                    {% endif %}
-                    {% if c.economic_interests %}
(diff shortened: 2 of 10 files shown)


theodormarcu:
Currently reviewing this.

theodormarcu:
@k-y-le I'm trying to review this but I'm dealing with an odd error. The lines point to the mixpanel code, but I removed it on my local copy and I didn't get any results.

**Pages affected:**

* Organization Map Page
Screen Shot 2020-05-25 at 12 45 43
* Organization Review Page
Screen Shot 2020-05-25 at 12 45 38
* Campaign Map Page
Screen Shot 2020-05-25 at 12 45 51
* Campaign Review Page
Screen Shot 2020-05-25 at 12 45 57

theodormarcu:
I did some digging and it seems that there is a problem with ```get_item```. More specifically, if the entry does not have an address associated with it, then the ```get_item``` template tag will not fail gracefully. I propose that this is addressed in this PR and you first check whether an entry has an address or not. It is possible that older entries will not have addresses.

Otherwise, I checked the visual updates and they are beautiful. I like how the address and name are displayed, although in the future we might have to consider making them more functional (but that's after we get more data!). I am such a big fan @k-y-le :)

lauren-johnston:
> I did some digging and it seems that there is a problem with `get_item`. More specifically, if the entry does not have an address associated with it, then the `get_item` template tag will not fail gracefully. I propose that this is addressed in this PR and you first check whether an entry has an address or not. It is possible that older entries will not have addresses, which is why this should be addressed now.
> 
> Otherwise, I checked the visual updates and they are beautiful. I like how the address and name are displayed, although in the future we might have to consider making them more functional (but that's after we get more data!). I am such a big fan @k-y-le :)

Worth trying `get_object_or_404()`: https://docs.djangoproject.com/en/3.0/topics/http/shortcuts/#get-object-or-404

theodormarcu:
Thanks for the suggestion! I think this might be appropriate in other instances as you want the page to still display the entry if it has no address associated with it. The problem isn't that the model raises a DoesNotExist exception (the model and queryset exist fine), but that the template expects that each community has an address associated with it and tries to show it. Thus, when an org admin/moderator looks at an older entry that does not have an address, Django attempts to get the address for it based on the entry id within the template, but that raises a Python KeyError.

*Theodor Marcu* | Princeton University | Computer Science | +1 (609) 933-5057 | theodormarcu.com ( http://theodormarcu.com )

lauren-johnston:
> Thanks for the suggestion! I think this might be appropriate in other instances as you want the page to still display the entry if it has no address associated with it. The problem isn't that the model raises a DoesNotExist exception (the model and queryset exist fine), but that the template expects that each community has an address associated with it and tries to show it. Thus, when an org admin/moderator looks at an older entry that does not have an address, Django attempts to get the address for it based on the entry id within the template, but that raises a Python KeyError.

Oh gotcha! Sorry I should've looked into this further before making the suggestion. In this case, it seems like doing a filter instead of a get would do the trick because that won't return an error. Then we can just check if the filter query returned an empty query set in which case we would not display anything.

k-y-le:
> I did some digging and it seems that there is a problem with `get_item`. More specifically, if the entry does not have an address associated with it, then the `get_item` template tag will not fail gracefully. I propose that this is addressed in this PR and you first check whether an entry has an address or not. It is possible that older entries will not have addresses, which is why this should be addressed now.
> 
> Otherwise, I checked the visual updates and they are beautiful. I like how the address and name are displayed, although in the future we might have to consider making them more functional (but that's after we get more data!). I am such a big fan @k-y-le :)

hm so the logic in the template tags is to check whether or not an entry has a phone number. if so, then it displays the phone number and address -- I don't think it's possible for any entries to have a phone number but no address associated with it? since both are required in the current form and were not in the previous one, unless i'm missing something there

i can push a fix but it would involve making a new template tag i'm pretty sure and just want to make sure that's necessary

(also finding it hard to test because i don't have any old entries in my local db and there's a non-null constraint on the phone number field so i can't delete it)

lauren-johnston:
> > I did some digging and it seems that there is a problem with `get_item`. More specifically, if the entry does not have an address associated with it, then the `get_item` template tag will not fail gracefully. I propose that this is addressed in this PR and you first check whether an entry has an address or not. It is possible that older entries will not have addresses, which is why this should be addressed now.
> > Otherwise, I checked the visual updates and they are beautiful. I like how the address and name are displayed, although in the future we might have to consider making them more functional (but that's after we get more data!). I am such a big fan @k-y-le :)
> 
> hm so the logic in the template tags is to check whether or not an entry has a phone number. if so, then it displays the phone number and address -- I don't think it's possible for any entries to have a phone number but no address associated with it? since both are required in the current form and were not in the previous one, unless i'm missing something there
> 
> i can push a fix but it would involve making a new template tag i'm pretty sure and just want to make sure that's necessary
> 
> (also finding it hard to test because i don't have any old entries in my local db and there's a non-null constraint on the phone number field so i can't delete it)

For flexibility (if we decide to make phone numbers optional for people not submitting to campaigns for example)  I think it would be helpful not to associate these two fields together.

k-y-le:
> > > I did some digging and it seems that there is a problem with `get_item`. More specifically, if the entry does not have an address associated with it, then the `get_item` template tag will not fail gracefully. I propose that this is addressed in this PR and you first check whether an entry has an address or not. It is possible that older entries will not have addresses, which is why this should be addressed now.
> > > Otherwise, I checked the visual updates and they are beautiful. I like how the address and name are displayed, although in the future we might have to consider making them more functional (but that's after we get more data!). I am such a big fan @k-y-le :)
> > 
> > 
> > hm so the logic in the template tags is to check whether or not an entry has a phone number. if so, then it displays the phone number and address -- I don't think it's possible for any entries to have a phone number but no address associated with it? since both are required in the current form and were not in the previous one, unless i'm missing something there
> > i can push a fix but it would involve making a new template tag i'm pretty sure and just want to make sure that's necessary
> > (also finding it hard to test because i don't have any old entries in my local db and there's a non-null constraint on the phone number field so i can't delete it)
> 
> For flexibility (if we decide to make phone numbers optional for people not submitting to campaigns for example) I think it would be helpful not to associate these two fields together.

ok makes sense - i'll add the template tag and make the change

k-y-le:
just committed that change @theodormarcu lmk if you're still getting the error

theodormarcu:
Checking now!

theodormarcu:
Works great for me. Thank you @k-y-le :) You're the best!
Screen Shot 2020-05-25 at 18 44 06

PR #138, opened 2020-05-25
Campaign/state content changes

lauren-johnston:
Key changes:
- Campaign and state page content changes

Screenshots:
Screen Shot 2020-05-25 at 4 55 05 PM
Screen Shot 2020-05-25 at 4 56 05 PM


psiyer7:
looks good!
Maybe can the 'Our campaign / Redistricting in MI' headers just be bold to break the text up a bit more?

Also, @k-y-le is it still possible to have the MI image on this campaign page? Think the images are awesome / makes it look more personalized and legit (asking just for this period/ we can figure out a more sustainable option later but might be an option that for every campaign we can include the image of the state its created in)

PR #139, opened 2020-05-26
Form Validation Updates

### LINKED ISSUE #103: Unclear error when invalid phone number
**Describe the bug**
If you enter in an invalid phone number on the new entry page, there is no error message and it's unclear where the error occurs (goes back to search part of page)

**To Reproduce**
1. Draw entry map
2. Include an invalid phone number (+61 on my australian number did it for me)
3. Press submit entry

**Expected behavior**
Highlight the phone number part of the form

### LINKED ISSUE #105: Entry Page on Failed Submit Does Not Highlight Census Blocks

### LINKED ISSUE #111: Better Form Validation

theodormarcu:
Closes #103. Closes #111. Closes #105.

A few form validation changes:

- If the user enters a polygon correctly but makes a mistake in another field the map will redraw the polygon, highlight the right blocks, and zoom into the right state. Uses sessionStorage to remember the state found by the user via the geocoder.
- Added form entry errors on all fields using alerts.
- Added red borders to fields in order highlight fields missed by the user.
- FIXES error on pressing "Draw" again. In the past, if user pressed draw, the polygon would disappear but that wouldn't update the entry. Thus, it was possible for the user to submit the wrong polygon if they didn't pay attention.

How to test:

- Try to leave a field non filled. The border should turn red.
- Draw a polygon but DONT fill any of the interests fields. This will trigger an error but will attempt to submit the form. Then you can observe the map behavior.
- Draw a polygon. Then press on "Draw polygon" again. Try to submit form. It should not let you.

One screenshot (hard to replicate other behavior):
Screen Shot 2020-05-25 at 23 21 19

### DIFF
--- a/main/static/main/css/style.css
+++ b/main/static/main/css/style.css
@@ -15,7 +15,15 @@ and rules, since this can lead to unintended consequences.
 This means no changing anything like <a>, <li>, <p>, etc.
 */
 
-input:invalid {
+form:invalid {
+  border-color: red;
+}
+
+.has_error {
+  border-color: red;
+}
+
+.has_success {
   border-color: red;
 }
 
--- a/main/static/main/css/style.css
+++ b/main/static/main/css/style.css
@@ -15,6 +15,10 @@ and rules, since this can lead to unintended consequences.
 This means no changing anything like <a>, <li>, <p>, etc.
 */
 
+input:invalid {
+  border: 1px solid red;
+}
+
 /* TOC Checkbox*/
 .toc-check {
   display: inline;
@@ -192,13 +196,13 @@ footer {
 /* Help dropup */
 
 #dropuphelp {
-  position: fixed; 
-  bottom: 40px; 
-  right: 40px; 
+  position: fixed;
+  bottom: 40px;
+  right: 40px;
   z-index: 1010;
 }
 
-.dropdown-menu a{
+.dropdown-menu a {
   color: #333;
 }
 
@@ -208,8 +212,8 @@ footer {
 /* Bootstrap Buttons. */
 
 #video_popup {
-  position: fixed; 
-  bottom: 90px; 
+  position: fixed;
+  bottom: 90px;
   right: 24px;
   z-index: 1000;
 }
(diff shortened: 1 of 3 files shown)


k-y-le:
Good changes! Messed around with drawing and stuff a lot, and almost everything acted as expected. The one thing that didn't work out is when I delete all the vertices of a polygon (well, all but two, but then the polygon disappears), then I am unable to draw on the map anymore. There are still the buttons for clear, edit, and finish drawing in the top right, but clicking anywhere on the map while trying to draw and nothing will pop up.

one other issue I had is that upon entering the entry page, the personal info fields were all already highlighted in red, despite not having entered anything yet

image
the same thing happens when i try to sign up
image

also a small detail and not necessarily in scope but it would be nice upon failure if it moved you immediately to the section with an error -- for instance when i left the interest fields blank i was borught right to the top and was a little confused what happened. either that or a small message saying something along the lines of "There was an error in submitting your community. Please check over and ensure that you have filled out all required fields." -- this is probably the easiest solution and would make sense for the user

theodormarcu:
I'll look into all the comments above shortly. Thanks for the attention to details y'all! It will be done by our meeting.

theodormarcu:
Addressed all your concerns.

PR #141, opened 2020-05-26
Form Visual Updates

theodormarcu:
- Updated spacing and text on address. It's common to add both placeholder text and normal text labels.
- Updated User Name -> Full Name (only front-end change, didn't touch model)
- Updated Community Name Example
Screen Shot 2020-05-26 at 00 30 51

Screen Shot 2020-05-26 at 00 30 44


k-y-le:
oh! just noticed but the numbering of the form is off haha - draw your community section should be 4 instead of 6

theodormarcu:
Good point! Fixing now.

PR #144, opened 2020-05-26
Added shepherd.js for map drawing

arorasomya:
- added steps to explain the controls on the map
- added button that would trigger the tour
- button only shows once the map is revealed

image
image
image
image
image
image
image

### DIFF
--- a/main/static/main/js/geo.js
+++ b/main/static/main/js/geo.js
@@ -965,8 +965,8 @@ let myTour = new Shepherd.Tour({
 
 
 myTour.addStep({
-  title: 'Community Drawing Tutorial',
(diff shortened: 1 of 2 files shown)


arorasomya:
> Looks really good! is it possible to have the shepherd js tutorial hide the buttons less by being to the side at certain points (for the draw polygon step for example). Sometimes it feels like the steps of the tour cover up the buttons they're trying to highlight. Not a big deal though.

For the draw polygon step, the only buttons visible on the map then are the draw polygon button and the show census blocks. For the other buttons - edit, delete, finish - I tried side in the beginning, but the right arrow doesn't show because of the white color - so that would need to include some design changes for the popup - like maybe an outline.

PR #167, opened 2020-05-29
visualization updates

k-y-le:
- added loading animation instead of text

- org/campaign maps now zoom to state instead of communities

- cleaned up landing page html + made dropdown smaller on mobile

- fixed collapse issue on viz dropdowns

- update select to standardize appearance


k-y-le:
> Looks really good! Map code looking really nice & really glad about the zoom to state functionality.
> 
> Requesting a small change on the map dropdown on front page. Text is not centered and the dropdown arrow covers it sometimes.
> Screen Shot 2020-05-29 at 11 04 46 AM
> 
> I think we could make the whole map component & the text in the dropdown a bit smaller which would probably help fix the issue of text being too big for the dropdown.

made these changes -- didn't center since that's not possible with an html select element (and may look a bit strange, imo) -- in the future we may want to turn it into a jquery span if we move to a different overall style

lauren-johnston:
Thanks for looking into this! Going to approve bc it's no longer a breaking issue but I think we do need to look into centering the text or padding because it looks too far to the left. This can be considered in a future PR though. Screenshot below for example: 
Screen Shot 2020-05-31 at 10 19 44 PM

Issue #165, opened 2020-05-28
Remove/change inconsistent placeholders on entry section 3


lauren-johnston:
**Is your feature request related to a problem? Please describe.**
Low priority - placeholder text on section 3 does not match other section.

**Describe the solution you'd like**
Placeholder text can be changed to the same format as others. "..." can be removed.

<img width="924" alt="Screen Shot 2020-05-28 at 11 04 47 AM" src="https://user-images.githubusercontent.com/8892509/83177044-4686bb80-a0d3-11ea-8ddb-9bdeff906214.png">

theodormarcu:
I think it's important to leave placeholder texts here because there is no "Show Examples" button.

lauren-johnston:
> I think it's important to leave placeholder texts here because there is no "Show Examples" button.

Ok gonna suggest that we reformat "My local community..." to remove the "...".  We could make the placeholder "Community Name".

theodormarcu:
Sounds good!

theodormarcu:
Closed via #179

Issue #163, opened 2020-05-28
Add Information About Census Blocks in Drawing Section


theodormarcu:
**Is your feature request related to a problem? Please describe.**
Users, including VNP, seem to be getting confused about the blue census block blobs.

**Describe the solution you'd like**
Include a statement about census block drawing in the instructions and in the map drawing section.

**Describe alternatives you've considered**
It seems that not telling them anything about drawing is not enough.

theodormarcu:
Closed by #179.

PR #179, opened 2020-06-02
Content Changes

theodormarcu:
Updated changes based on conversations with our partners.

1. **Entry page** is updated per our internal conversation.
a. Added instructions to get started.
b. Changed "Personal Information" to "Identifying Information"
c. Clarified instructions for filing interests fields.
d. Removed "Recommended" tags.
e. Changed the order of interests fields.
d. Removed issues (front-end ONLY for now). Discuss during meeting.
e. Updated instructions for drawing community.
f. Changed "vertex" to "point" on front-end.
g. Added disclaimer about editing communities on entry page.
h. Renamed "Save Entry" to "Save Community".
2. Updated **Thanks Page**.
a. Added updated disclaimer.
b. Added "Send Feedback" button and section.

Screenshots:
Screen Shot 2020-06-02 at 15 13 40

Screen Shot 2020-06-02 at 15 15 02
Screen Shot 2020-06-02 at 15 15 11

Screen Shot 2020-06-02 at 15 14 50

Screen Shot 2020-06-02 at 15 13 11

### DIFF
--- a/main/forms.py
+++ b/main/forms.py
@@ -119,7 +119,7 @@ class CommunityForm(ModelForm):
             "census_blocks_polygon_array": forms.HiddenInput(),
             "census_blocks_polygon": forms.HiddenInput(),
             "entry_name": forms.TextInput(
-                attrs={"placeholder": "My local community..."}
+                attrs={"placeholder": "Community Name"}
             ),
             "entry_reason": forms.Textarea(attrs={"rows": 5}),
             "cultural_interests": forms.Textarea(attrs={"rows": 5}),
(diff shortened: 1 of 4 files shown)


theodormarcu:
Thank you for the review! Right now I changed so that you need to add 1/4 of the questions (other takes the place of issues). So you could complete say other and still submit the form. I thought this was the best way because it was most flexible. Do you think this will be a problem?

I will add a separate issue for #2. Not sure if we can easily change that for now.

lauren-johnston:
> Thank you for the review! Right now I changed so that you need to add 1/4 of the questions (other takes the place of issues). So you could complete say other and still submit the form. I thought this was the best way because it was most flexible. Do you think this will be a problem?

Sounds good!

> I will add a separate issue for #2. Not sure if we can easily change that for now.

Great! Agreed it seems out of scope.

theodormarcu:
Closes #165.

theodormarcu:
Closes #163.

PR #184, opened 2020-06-05
export pdf from submission page

k-y-le:
you can now export your community map + info as a pdf!

- added a button on submission page for exporting pdf which on click: gives loading popup on map, centers map on community (in case not already), and creates a pdf which is saved

pdf includes
- name of community
- link to Representable community page
- org/campaign (if applicable)
- image of the map:: this is taken from the canvas, so can be manipulated by the user ! -- that's why it currently delays creating the pdf, so that there is time for the map to zoom to the community's bounding box, if necessary. This also means that if anyone wants to include state legislature/counties/census block lines, all they have to do is select them and they will be included in the pdf. (ongoing q should be: is it necessary to include/not include any of these lines? is it confusing if someone selects those lines and then sees them on the pdf?)
- whatever community info is included from the five categories

the pdf isn't the prettiest thing but i think it serves a good purpose for now! + formatting is hard thru jspdf (which has probably the worst documentation i have ever seen lol)

ALSO of note: if you click on the export pdf button too fast before the map loads, it won't give a very good image, since it will still be zooming in when the image is created. i dont foresee this being an actual issue tho!

🤠 screens 🤠 

comm w org+campaign:
image
comm w/out org+campaign, with all data layers selected:
image
and the community info (on 2nd page):
image


k-y-le:
another thing i forgot to add that's a todo -- in michigan feedback we learned that having identifying info with the exported community would be useful, but currently anyone can view the submission page. so, on the todo list is including the personal info, but only if the person viewing the submission is the person who created it! (might need a disclaimer somewhere in that case too) 🦞 🏭

lauren-johnston:
Awesome! Will take a look tomorrow.

PR #196, opened 2020-06-10
Upgrade to Django 3.0.7

lauren-johnston:
***Key changes:***
- Upgrades to Django 3.0.7 and upgrades all packages alongside it. Eventually, support for Django 2.2 will be dropped and third party libraries may stop supporting Django 2.2. In addition, security fixes will likely be focused on most recent versions of Django. Therefore, if we can upgrade now it's a good idea. 
- Removes unnecessary packages that were in requirements.txt.

***Steps to test:***
Create a new virtual env (if you need to remove the old one, run rm -rf venv):
```
python3 -m venv venv
. venv/bin/activate
```
Install packages:
```
Run pip install -r requirements.txt
```

Install precommit in the venv:
```
pre-commit install
```

Run the server and navigate to different pages. Check that the map loads. Check that you can create an entry and that form is rendered properly.


lauren-johnston:
Unfortunately it seems like Travis CI has outdated geo libraries so checks are failing: ```AttributeError: /usr/lib/libgdal.so.1: undefined symbol: OGR_F_GetFieldAsInteger64```. Not sure how to fix this.

lauren-johnston:
Closing for now, to be reviewed in future.

PR #198, opened 2020-06-10
Integration testing suite scaffolding

indraneel:
This is a barebones set up using [Cypress.io](https://cypress.io) for integration testing

To get it working you'll either have to:

- download the Cypress client https://download.cypress.io/desktop 
- run the Django App and DB
- run the Cypress testing suite using the above Client

OR

- follow the instructions to get Cypress working with Docker https://github.com/bahmutov/cypress-open-from-docker-compose I was able to do it but I cannot find the code atm and I am afraid I deleted it. But, it is definitely doable

### DIFF
--- a/cypress.json
+++ b/cypress.json
@@ -1,0 +1,4 @@
+{
+    "baseUrl": "http://localhost:8000"
+  }
+  
--- a/cypress/integration/spec.js
+++ b/cypress/integration/spec.js
@@ -1,0 +1,8 @@
+describe('logs in', function() {
+    it('visits homepage', function() {
+        cy.visit('/');
+    
+        cy.get('.row-hero')
+            .should('contain', "We're creating maps of communities to end gerrymandering.")
+    })
+})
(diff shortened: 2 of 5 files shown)


lauren-johnston:
We can discuss including Cypress in scope for the next sprint (starting Tues).

lauren-johnston:
It seems like our testing needs right now are covered by the Django testing suite (in tests.py) and we've got a lot of work ahead of us there to add more tests. So I think I'm going to put this on hold for now - at least until we make substantial progress on adding tests. Really appreciate the PR-- just not sure if it can help us immediately & don't want to add an extra tool yet when tests.py seems sufficient for our purposes.

lauren-johnston:
though @indraneel happy to reopen if you think cypress is going to be substantially easier for frontend testing - it's hard for me to say

indraneel:
yeah i don't think it makes sense to worry about testing when we're busy w/ features! we can revisit it later. 

for some context in case it helps thinking about this, cypress is for integration tests — basically automating the full website experience similar to how selenium works. so this goes beyond frontend unit testing, it is best used for critical application paths that touch frontend and backend features. basically, to prevent this: https://thumbs.gfycat.com/BriefUniformHornet-size_restricted.gif

lauren-johnston:
Ooh that makes sense, that sounds like it will be really helpful in future, especially once we've got some basic unittests. We can revisit again later in the summer.

PR #240, opened 2020-06-22
geojson + pdf export now contain identifying info if user has permission

k-y-le:
On the submission page, if the user is admin/moderator of the entry organization or the creator of the entry being viewed, it now displays identifying info (name + address)

This is also included when exporting -- visible on the pdf and in the geojson data

to test:
- go to submission page for any submission you created / you are org admin for
- check that identifying info is visible
- export pdf/geojson and check identifying info is there
- repeat for community you should /not/ have access to identifying info for

screens:
image
image


lauren-johnston:
Looks great! Going to review when I get a chance today after these upcoming meetings!

lauren-johnston:
Also adding @theodormarcu  to this to maybe take a look... since this is PI related I want to make sure we uncover any possible issues before deploying.

k-y-le:
> I tested and found that the permission control works as planned at least for displaying info on pages! But I'm running into this error, unfortunately, even after running make migrations (this is on the organization map page and the campaign map page after clicking on a community):
> 
> Screen Shot 2020-06-22 at 10 50 47 PM

Good catch! Looks like this occurs if there are two organizations in the db with the same name (i was calling get() by org name) I'm committing a change to query by slug, now :: glad you caught this

We may want to change it in the future so that organizations cannot have the same name? but it's not a priority

theodormarcu:
I think generally it's good to query by the UUID/unique identifier. We allow orgs to have the same name just in case, and it's a good practice for this kind of stuff.

main/views/main.py
```diff
@@ -225,13 +226,27 @@ def get(self, request, *args, **kwargs):
...
+            "city": city,
             "c": user_map,
             "entries": json.dumps(entryPolyDict),
             "mapbox_key": os.environ.get("DISTR_MAPBOX_KEY"),
             "mapbox_user_name": os.environ.get("MAPBOX_USER_NAME"),
         }
+        if self.request.user.is_authenticated:
+            if user_map.organization:
+                context["is_org_admin"] = self.request.user.is_org_admin(
+                    Organization.objects.get(slug=user_map.organization.slug).id
```
lauren-johnston, line 244: This can be changed to user_map.organization__id I believe - no need to query again!

theodormarcu: Good point!

main/templates/main/pages/submission.html
```diff
@@ -32,7 +32,7 @@
     <div>
       <script type="text/javascript">
         let a = '{{ entries | escapejs}}';
-        var mapbox_user_name = "{{mapbox_user_name}}";
+        let mapbox_user_name = "{{mapbox_user_name}}";
```
lauren-johnston, line 35: hmm.. this would display our mabox_user_name to all users - do we need to include?

k-y-le: i'm not totally sure about this one.. we've been using this line since pretty much the beginning. unless you mean the change from var to let? i don't know that it changes things, but i'm not totally clear on this aspect of things (imagine if we were taught javascript at princeton...)

k-y-le:
> Hey Kyle! Changes above look great. I noticed that there's an error now when opening submissions from the public map page, sorry to keep coming back to you with change requests:
> 
> Screen Shot 2020-06-23 at 3 21 45 PM

Another good catch.. i definitely should've tested more! Pushed a fix

PR #243, opened 2020-06-23
Added mandatory verification for dev branch

### LINKED ISSUE #242: Make email verification mandatory on dev
It's useful for developers to be able to replicate the sign-up flow locally so they can double-check the user flows.

- Make email verification mandatory on dev.
- Use SMTP server for dev as well as prod settings.

- We tried making it optional but some inconsistencies can creep in.
- It's also hard for developers to replicate the setup.

**Additional context**
- Edward and Theodor spoke about this and discovered it when talking about #234

theodormarcu:
Co-Authored-By: Lauren Johnston <lauren-johnston@users.noreply.github.com>

Closes #242 

**Description of changes:**
- Made email verification mandatory on dev.

**To replicate:**
1. Start a local server.
2. Create a new account. 
3. Check that you received a verification email and that the account was approved. 
4. Create a test community.


theodormarcu:
I was. Edward is having similar issues. I am pushing a commit that uses the terminal-based email backend on dev in order to avoid this problem.

theodormarcu:
Updates:
- Email verification mandatory on dev *for now*.
- Dev email verification uses console.
- Implemented custom Django All auth adapter for debugging that prints a simpler message to the terminal ('main/users/adapter.py').

Screen Shot 2020-06-25 at 12 31 26

theodormarcu:
@ETedward This might be of interest to you. In order to print a simpler message I had to implement a custom email adapter, which could also be used in production to change the behavior of logging in after confirming email.

representable/settings/prod.py
```diff
@@ -10,14 +10,16 @@
 
 DEBUG = False
 
-ACCOUNT_EMAIL_VERIFICATION = "mandatory"
-
-ACCOUNT_EMAIL_CONFIRMATION_EXPIRE_DAYS = 7
+# Production Email Settings
+EMAIL_BACKEND = "django.core.mail.backends.smtp.EmailBackend"
```
lauren-johnston, line 14: is the EMAIL_BACKEND line necessary in prod? I don't see it in the version that's currently live

theodormarcu: It's not but I think it's useful to be explicit here (it's the default). I often find myself wondering whether by removing it I break the system.

PR #247, opened 2020-06-25
State model

### LINKED ISSUE #237: Create model,urls for U.S. States (e.g. r.org/states/mi)
Context:
We need pages with specialized content for each state. This should be like http://127.0.0.1:8000/michigan/ but editable from the admin, preferably with rich text editing

jessiejofielding:
Closes #237 

**Description of changes:**

- create State model
- create State admin view with rich text editor form
- create a temporary/test state.html state template
- set up URLs (/state/mi/ or /state/<any state 2 letter abbreviation>/)
- migrations
- update requirements.txt

**To replicate:**

1. Install requirements.txt
2. python manage.py migrate
3. Start a local server
4. Open admin page in browser
5. Under "Main", click on "State"
6. Add a state, and save it (when adding a state, the content field form should be a rich text editor)
7. Navigate to /state/<your_state_abbreviation> - this layout/template is not intended to be final, and displays primarily the content field of the state. Until states are added to the database, this will redirect to "/", so we can ensure these pages remain unreachable until we have finalized a template.

image


lauren-johnston:
Amazing! looks really good. I know I'm not one of the auto-selected reviewers but wanted to give some quick feedback.

What do you think about prepopulating db with fixtures to setup state pages? Just to create a default state page that can be customized + so everyone can just load the fixture w/o creating each state on their local env https://docs.djangoproject.com/en/2.2/howto/initial-data/

k-y-le:
Actually, in hindsight, I think it would be better to combine the two, since we may not always want to include a section like "redistricting news" depending on the state

jessiejofielding:
Since Kyle is making field changes on his branch, I have left them as is here, and just included the state fixture json and the script to rewrite that file with any generic state content. To use (if you don't want to adjust the fixture, skip to step 3): 

1.  edit create_state_fixtures.py to include 
`field["<field_name>"] = <desired_content>`
on line 20.

2. run python create_state_fixtures.py

3. run python manage.py loaddata states.json

PR #248, opened 2020-06-25
[Draft] gmail-based email verification

lauren-johnston:
Edward and Theo - not sure if this is super helpful to y'all but since Sendgrid has been throttling emails, I figured it might be best to just send emails straight from gmail - it seems that many people have used gmail as a way to avoid paying for email verification before anyways.

Key changes:
- Uses gmail to send email rather than send grid
- Moves all email settings into base.py (we would verify the same way in prod vs local)

To replicate:
- First, add our gmail password to your .bash_profile. You can find the gmail password in our Bitwarden.

Run nano ~/.bash_profile

Add the following line:
```
export GMAIL_PASS="..."
```

- Next, create an account using a valid email. 
- See that a real email appears in your inbox and click on the verification email.
- Sign in with your verified account.

### DIFF
--- a/representable/settings/base.py
+++ b/representable/settings/base.py
@@ -234,6 +234,8 @@ elif DATABASES["default"]["ENGINE"] == "django.db.backends.sqlite3":
         "ENGINE"
     ] = "django.contrib.gis.db.backends.spatialite"
 
+ACCOUNT_AUTHENTICATION_METHOD = "username_email"
+
 ACCOUNT_EMAIL_VERIFICATION = "mandatory"
 ACCOUNT_EMAIL_REQUIRED = True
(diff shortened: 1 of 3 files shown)


theodormarcu:
I couldn't get it to work with my Princeton account. Worked well with my Gmail. :(

ETedward:
Thanks Lauren! this is super helpful with testing emails,
Both Princeton and Gmail works for me.

lauren-johnston:
Closing this PR for now - we started paying for SendGrid and found that this fixed the issue with our emails not being delivered (the free ip address was blacklisted by some providers). so Gmail is no longer necessary at the time being!

PR #250, opened 2020-06-26
redirecting to initial link after signup

ETedward:
version where redirect to entry or campaign pages works when ACCOUNT_EMAIL_VERIFICATION is set to "optional"
tested a few cases: when user logins directly; when user logins unsuccessfully (ie. with wrong password) then logins; when user creates new account; when user creates new account unsuccessfully (ie. username already exists) then creates new account.


lauren-johnston:
If it's helpful, these are the errors on the signup page that cause the redirect so you can try to replicate: 

Screen Shot 2020-06-25 at 7 30 19 PM

ETedward:
1. For the redirect after a failed login and sign up, after some testing, it seems that AllAuth was able to automatically handle those redirects. Is that able to work for you locally? 
2. Yes! with the "optional" we can work on adding a warning / re-verify back into the flow after someone has submitted a community.

theodormarcu:
I think that should be included in this PR. I know PRs are supposed to represent atomic changes, but this seems to be tightly coupled.

lauren-johnston:
Closing this for now so we can change the "master" branch to "main".

PR #257, opened 2020-06-29
state page template

k-y-le:
based on #247 (shoutout to jessie! looks great 🕺 )

**changes**
- updated model rich text fields to be slightly more descriptive + combined first two
- integrated most of the figma designs for the state page template, as closely as is feasible for now
- added campaigns within the state page + links

**to test**
1. python manage.py migrate
2. run server
3. log in to admin
4. create state with rich text fields
5. go to localhost/state/{abbreviation}

**please note**
- responsiveness isn't perfect rn, will update as we finalize models/design/format
- i added campaigns as a simple table for now rather than as a carousel, since we still need to change the models so that campaigns / orgs can submit logos + i wanted to get this out now and start discussing. how will this affect the db? @lauren-johnston @jessiejofielding 
- currently all campaigns are displayed. we'll have to scope out how to implement / necessity of public vs private campaigns!!
- because styling svgs is kinda nasty, i had to add the fill color manually in the svg -- this is easy but will have to automate for every state later. as of right now michigan is the only state that's blue

**screens**
image

image


k-y-le:
oh i didnt realize this would include jessie's commits.. i have underestimated github once again. please only pay attention to mine! lol

PR #277, opened 2020-07-06
Simplify org dashboard permissions

### LINKED ISSUE #276: Simplify organization permissions to admin-only group
Organization moderators and members have no clear purpose at this point. Furthermore, MI partners and others want a hands-off approach to moderation and do not need a separate class of moderators and admins. This effectively replaces #224 . 

Reduce scope to admins; replace admins with permissions group.

Django guardian (too complicated for our needs right now)

**Additional context**

lauren-johnston:
Key Changes:
- Simplifies the Org dashboard permissions -> Admins only 
- Users no longer added from Allowlist to Member list 

The changes made above make it simpler to set dashboard permissions and test them. With only admins having access to the dashboard, we don't have to worry about member or moderator permissions - they'll just receive a "403 forbidden" on pages if they were to try to access. 

Tests Added:
- Added tests for organization admins vs. members to demonstrate that only admins can view private dashboard pages.
- Membership form defaults to is_org_admin = True

To Replicate:
- Migrate
- `python manage.py test`
- Visit dashboard as admin

Screenshots:
Screen Shot 2020-07-06 at 11 12 09 AM
Screen Shot 2020-07-06 at 11 12 17 AM
Screen Shot 2020-07-06 at 11 12 22 AM

resolves #276


lauren-johnston:
Also - I would like to figure out if there's a better way to do permissions (permissions group etc) but need to figure out a custom migration for that.

lauren-johnston:
> Is it intentional that we currently don't allow a way to change permissions from admin to not-admin? If so, should we remove language that suggests you can (mainly so that people are aware when they add admins that they can't later remove them)?

You're right - I can remove that language for now then add a way to remove people next week

Issue #284, opened 2020-07-09
Highlight all census blocks/block groups intersected by community polygon (even partially)


lauren-johnston:
**Is your feature request related to a problem? Please describe.**
Census blocks change between census years so asking community drawers to draw their community around old census blocks may be misleading. In addition, similar drawing tools that highlight census units (such as this one https://medium.com/nyc-planning-digital/building-a-custom-draw-mode-for-mapbox-gl-draw-1dab71d143ee) implement this inclusive drawing style. 

**Describe the solution you'd like**
Change tool to inclusive selection

**Describe alternatives you've considered**
Custom Draw tool instead of polygon

**Additional context**
<img width="853" alt="Screen Shot 2020-07-09 at 11 49 47 AM" src="https://user-images.githubusercontent.com/8892509/87079270-a4cab200-c1da-11ea-8895-f23eecab6ba1.png">

<img width="836" alt="Screen Shot 2020-07-09 at 11 52 03 AM" src="https://user-images.githubusercontent.com/8892509/87079256-9f6d6780-c1da-11ea-9152-c3a37a2b775b.png">

Issue #279, opened 2020-07-07
Move to block groups as basic unit of mapping


k-y-le:
**Is your feature request related to a problem? Please describe.**
As per team discussion, we are moving to block groups as the basic unit of mapping, which will hopefully resolve a few of the bugs users currently experience related to census block loading

**Describe the solution you'd like**
Drawn polygons snap to census block groups

**Describe alternatives you've considered**
Census tracts, blocks, precincts, no units at all

**Additional context**
Related to discussions about drawing tools (need to scope further)

PR #287, opened 2020-07-10
Block group snapping for key states

k-y-le:
**changes**
- on entry page, view census block group lines, and highlight them when drawing a polygon (#279 )
- block group files for key states uploaded to mapbox: MI, VA, IL, NJ, TX, PA (waiting to upload others until we decide how best to move forward here + hear from mapbox)
- drawing a community now highlights all polygons which have overlap with the user polygon, rather than just those entirely contained by the polygon (#284 )
- saving polygons without querying census layers no longer gives errors
- removed references to census blocks (button on map page, stop on shepherdjs tour)
- removed geocoded point when searching (clutter, imo)

**to test**
- go to entry page
- try drawing a community in key states
- play around with tools to try and break it
- try drawing in other states & confirm no errors

**screenshots**
before drawing:
image
after:
image


lauren-johnston:
Reviewing now! Exciting stuff - thanks for opening this. Before we merge, we may want to think about using Celery or something to migrate the old polygons for census blocks to the block groups. Otherwise, we'll have a mismatch in old polygons vs. new.

k-y-le:
> So I've been reviewing and so far tried MI, VA, and NJ. However, I'm seeing an "uncaught html error validate is not a function" right when the form submits. I tried to catch a screenshot but it switches really quickly.
> 
> Also, there's a weird thing where I can draw a small polygon that doesn't highlight any blocks:
> 
> Screen Shot 2020-07-13 at 3 28 51 PM
> 
> Finally, minor style suggestions:
> 
> These states look like different colors (may just be my imagination).
> 
> Screen Shot 2020-07-13 at 3 17 48 PM
> 
> Maybe we could draw the polygon not filled in? Or maybe just a different color palette so that the filled polygon is a darker shade of the same color or something. Feel free to disregard this - not a huge priority at all.
> 
> Screen Shot 2020-07-13 at 3 19 32 PM

looks like the validate error is something which has already been happening, and is unrelated to these changes (happens on main branch currently)

good catch on the small community bug ! i can't determine what causes it -- think that we can probably dismiss for now tho as we move to the select tool

the difference in color between michigan and illinois is a quirk of mapbox rendering -- basically because block groups in michigan cities tend to be less uniform in shape, mapbox's algorithm combines features at one higher zoom level for those places compared to cities like chicago (at least, that's my interpretation) -- so i don't think there's anything to be changed there

image
I updated the colors/opacity -- i think this is more clear!

lauren-johnston:
Awesome, thanks for making these changes!

I think the one thing we need to figure out now, per my first comment above is migrating old polygons:

> Before we merge, we may want to think about using Celery or something to migrate the old polygons for census blocks to the block groups. Otherwise, we'll have a mismatch in old polygons vs. new. 

We also may want to discuss in the sync tomorrow to what we want to do for states where census block groups are not loaded. For example, if I tried to draw in WA right now then view on map, there would be no polygon to show right? Or would it fallback to default. 

Finally, to the point about migrating old polygons, Chicago wants census blocks rather than block groups right? so we'll need to have a way to switch between the two. 

Sorry to delay this a bit with these questions, just want to be a bit careful since this is a substantial shift.

k-y-le:
> I think the one thing we need to figure out now, per my first comment above is migrating old polygons:
> 
> We also may want to discuss in the sync tomorrow to what we want to do for states where census block groups are not loaded. For example, if I tried to draw in WA right now then view on map, there would be no polygon to show right? Or would it fallback to default.
> 
> Finally, to the point about migrating old polygons, Chicago wants census blocks rather than block groups right? so we'll need to have a way to switch between the two.
> 
> Sorry to delay this a bit with these questions, just want to be a bit careful since this is a substantial shift.

True, let's talk tomorrow! 

Currently, for states which are not covered by blockgroups, the user polygon (no snapping) will be saved

And on Chicago, we're meeting with them on Wednesday! Won't be sure about what exactly their needs are until then.

lauren-johnston:
requesting @theodormarcu as discussed in the meeting today :)

theodormarcu:
Here's what I found:

- Is it possible to disable polygons in states that are not active? This could happen, and it might look worse in neighboring states. I suggest we clear the map and reload the polygons in the specific state anytime the user uses the geocoder.
Screen Shot 2020-07-15 at 10 11 32

A better example of this is at the PA/NJ border. It's confusing when you live in a community on the border of a state:
Screen Shot 2020-07-15 at 10 13 05

In the above image you can also see a JS error, although I wasn't able to reproduce it when going from state to state in the same session.

If I go to state X, and then I use the geocoder to go to state Y and move the polygon/draw a new one in the same SESSION (without refreshing page), this is what I get (old state block highlighting never disappears):
Screen Shot 2020-07-15 at 10 16 02

Tested VA, NJ, PA, IL, TX, MI, and all census block groups work there.

k-y-le:
Thanks for the feedback Theo!
> Is it possible to disable polygons in states that are not active? This could happen, and it might look worse in neighboring states. I suggest we clear the map and reload the polygons in the specific state anytime the user uses the geocoder.
There's no easy way to disable polygons based on state, as far as i know (mapbox doesn't store what state you are in) -- but it would be possible to disable the block group lines (if that's what you meant) -- i'll add that!
> If I go to state X, and then I use the geocoder to go to state Y and move the polygon/draw a new one in the same SESSION (without refreshing page), this is what I get (old state block highlighting never disappears):
I'm also pushing a commit to clear the map when someone uses the geocoder to search for a location in a different state

hopefully this makes sense, basically the way it works now is:
- block groups only loaded for state you searched for
- searching for a location in a different state: clears block groups for the other state & clears the polygon

k-y-le:
also, the way this will work with the select tool is that one will not be able to draw a community that crosses state lines (when currently one can but only the blocks for the searched state will show up)

theodormarcu:
> There's no easy way to disable polygons based on state, as far as i know (mapbox doesn't store what state you are in) -- but it would be possible to disable the block group lines (if that's what you meant) -- i'll add that!

That works great!

> I'm also pushing a commit to clear the map when someone uses the geocoder to search for a location in a different state
> hopefully this makes sense, basically the way it works now is:
> block groups only loaded for state you searched for
> searching for a location in a different state: clears block groups for the other state & clears the polygon

Tested and works great.

PR #295, opened 2020-07-13
state template formatting updates

k-y-le:
**changes**
- responsiveness of state svg (wrapped in card)
- removed exclamation point
- jumbotron w/ rounded edges for COI section
- join drive button larger + primary coloring
- reduced drop shadow of cards

**to test**
1. go to state page
2. play w responsiveness

**of note**
- please comment any additional thoughts on formatting -- this is all easy stuff to change (for the most part)
- in the future, i think it would be great to have some themed drawings of some sort, like we have on the landing page
- maybe we want to de-emphasize the final "draw my community" button by making it an outline

**screenshot**
image

### DIFF
--- a/main/static/main/css/style.css
+++ b/main/static/main/css/style.css
@@ -444,6 +444,7 @@ color scheme. See readme.md*/
 .state {
   max-width: 70%;
   height: 50vh;
+  margin: auto;
 }
 .state-sm {
   width: 2rem;
@@ -854,7 +855,7 @@ label {
 .card-signin {
   border: 0;
   border-radius: 1rem;
-  box-shadow: 0 0.5rem 1rem 0 rgba(0, 0, 0, 0.1);
+  box-shadow: 0 0.3rem 0.6rem 0 rgba(0, 0, 0, 0.1);
 }
 
 .card-signin .card-title {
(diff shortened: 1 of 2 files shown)


okusio:
I would 
- Make the COI box smaller (smaller margins for the text)
- Center the community mapping drives heading/text
- i feel like there should be some difference between the buttons for join drive and draw my community, i think i agree with deemphasizing the draw my community button. During the user interview i had, the first thing he went to was draw my community instead of even looking at the fact that drives were up. I think a lot of people will go through this page very quickly.

k-y-le:
> I like that Join drive is more prominent now!
> 
> I think the state svg may be better not in a card? Especially bc it now resizes based on content so could be really large depending on left panel or really small like this:
> Screen Shot 2020-07-13 at 3 37 33 PM
> 
> Would recommend that the svg of the state be inline text wrapped. Rather than a box.
> 
> Also, not sure that svg in the header on mobile adds to the design - it doesn't line up exactly with the text:
> 
> Screen Shot 2020-07-13 at 3 36 58 PM

I added the card because of something Jessie had noted, that the design isn't particularly balanced, especially when there's more initial text. the screen here doesn't really apply bc i don't think there will be any case where we have no text on the state page

I'm not clear on how inline text wrapped svg will be any different from the current way of displaying the svg

And the mobile was just bc the state svg adds clutter, but otherwise would be entirely a wall of text. it's easy to remove, though

PR #296, opened 2020-07-13
Privacy and Tracking Consent Banner

theodormarcu:
* Added a privacy and tracking consent banner.
* Updated Privacy policy to reflect the third-party tracking services we use.
* Added an account settings pane in the dashboard (can be changed in the future, this is the closest we've got to an account page).

Screen Shot 2020-07-13 at 17 51 12

**How does tracking consent work?**

- Mixpanel is now disabled by default. Google Analytics will be auto-enabled, so we can track unique page visits.
- Users will be presented with a banner that nudges them to opt-in or opt-out from tracking on their first visit. If they select an option, the banner will not show again and the state will be saved in the browser's `localStorage`.
- Users can update their preferences by pressing on "Tracking Settings" in the footer or on their dashboard page.
- This PR only implements a client-side settings model. This means that if the same user visits the website from different browsers, they will have to Opt-in/Opt-out by again.

**Why client-side?**
This is the most basic and common way of saving state about our users. In the future, we might also need to implement a server-side account setting for logged in users, but that's a lot of work for now, and the payoff is minimal (logged in users will be the only ones to benefit).

**To test this PR:**
1. Open an incognito window.
2. Open the browser's console.
3. Type `localStorage.getItem("user-consent");` Return should be null.
4. Approve tracking in the banner.
5. Type `localStorage.getItem("user-consent");` in the console. Return should be true.
6. Go to footer. Press on "Tracking Settings". Opt out of tracking.
7.  Type `localStorage.getItem("user-consent");` in the console. Return should be false.
8. Log in. Go to the account dashboard.
9. Check that the state is "opted out".
10. Toggle the state a few times and check the console.

**Additions:**

**Privacy and Tracking Notice**
Screen Shot 2020-07-13 at 17 51 12
**"Tracking Settings" footer menu option. On click, it shows the banner again.**
Screen Shot 2020-07-13 at 17 51 18

**"Account Settings" dashboard pane. It includes a "Reset Password" button and a menu that displays the state of your tracking settings. This can be expanded in the future to accommodate server-side tracking.**

- Opt-in state.
Screen Shot 2020-07-13 at 17 51 28
- Opt-out state.
Screen Shot 2020-07-13 at 17 51 32

**Privacy Policy** The privacy policy now mentions Mixpanel and Google Analytics.

Screen Shot 2020-07-13 at 17 56 20


lauren-johnston:
ooh I love the new dashboard pane! 

the PR works for me. however, minor style things I noticed:
- users aren't guided to opt in. other sites I've seen only have one button "accept" or a small x in the corner to opt out. 
- doesn't blend in with the page. I think a light gray or white could be better for this. also we could make it smaller by removing the last line.
- feels like it takes up a lot of space on the page: could remove the line "by continuing to use ...." to make it smaller. This is especially noticeable on mobile.

Screen Shot 2020-07-13 at 3 57 39 PM

Otherwise looks good for me. I guess the one thing that's confusing is that GA is auto opt in vs. mixpanel is not. users may not expect this behavior.

theodormarcu:
> users aren't guided to opt in. other sites I've seen only have one button "accept" or a small x in the corner to opt out.

It's up to us how much we want to push it. I'm fine with guiding to opt-in.

> doesn't blend in with the page. I think a light gray or white could be better for this. also we could make it smaller by removing the last line.

Check out this version:

Screen Shot 2020-07-14 at 11 20 13
Screen Shot 2020-07-14 at 11 20 02

> Otherwise looks good for me. I guess the one thing that's confusing is that GA is auto opt in vs. mixpanel is not. users may not expect this behavior.

Check the updated language in /privacy.

> feels like it takes up a lot of space on the page: could remove the line "by continuing to use ...." to make it smaller. This is especially noticeable on mobile.
 
This is fine with me. Wanted to be as comprehensive as possible, like other websites do (e.g. Stack Overflow, example below), but ours does look indeed a bit bloated.

Screen Shot 2020-07-14 at 10 51 40

theodormarcu:
> Looks amazing, thanks theo :)

Thank you!!

PR #307, opened 2020-07-21
Multiple community geojson

### LINKED ISSUE #230: Organizations need a way to export multiple communities at once in GeoJSON
Context:
Organizations need a way to export multiple communities at once. Currently, we're able to export multiple communities from the Django admin side and share directly with partners in JSON and XLS (note that these formats are not valid geoJSON from what I understand).

Question: can we use the Django admin code that exports communities directly in the organization dashboard somehow?

jessiejofielding:
**Changes:**
- moves geojson logic in views/main.py from ExportView to its own function
- creates MultiExportView in views/partners.py to export a geojson file that is a FeatureCollection of all the communities (which are FeatureCollections themselves)
- creates URLs to MultiExportView - one for organization map with drive, one for organization map without drive
- adds appropriate link to the partner map context
- adds button to partner map page to export geojson

**To test:**
- run server
- go to an organization map page and click "Export geoJSON"
- go to a drive map page and click "Export geoJSON"

Closes #230

### DIFF
--- a/main/templates/main/partners/map.html
+++ b/main/templates/main/partners/map.html
@@ -47,8 +47,8 @@ integrity="sha384-MCw98/SFnGE8fJT3GXwEOngsV7Zt27NXFoaoApmYm81iuXoPkFOJwJ8ERdknLP
                 <p class="font-weight-light text-center">{{ drive }} </p>
                 {% endif %}
               </div>
-              <div>
-                <a class="btn btn-outline-primary btn-canvas mx-4" href={{multi_export_link}} role="button" download="communities.geojson">Export GeoJSON</a>
+              <div class="row justify-content-center">
+                <a class="btn btn-outline-primary btn-canvas mx-auto" href={{multi_export_link}} role="button" download="communities.geojson">Export GeoJSON</a>
               </div>              
               <div id="headingOne">
                 <button class="viz-dropdown btn btn-light font-weight-light" type="button" data-toggle="collapse" data-target="#collapseOne"
--- a/main/templates/main/partners/map.html
+++ b/main/templates/main/partners/map.html
@@ -124,6 +124,9 @@ integrity="sha384-MCw98/SFnGE8fJT3GXwEOngsV7Zt27NXFoaoApmYm81iuXoPkFOJwJ8ERdknLP
                   <tr>
                     {% endfor %}
                   </table>
+                  <div>
+                    <a class="btn btn-outline-primary btn-canvas mx-4" href={{multi_export_link}} role="button" download="communities.geojson">Export GeoJSON</a>
+                  </div>
                   {% endif %}
                 </div>
               </div>
--- a/main/templates/main/partners/map.html
+++ b/main/templates/main/partners/map.html
@@ -47,6 +47,9 @@ integrity="sha384-MCw98/SFnGE8fJT3GXwEOngsV7Zt27NXFoaoApmYm81iuXoPkFOJwJ8ERdknLP
                 <p class="font-weight-light text-center">{{ drive }} </p>
                 {% endif %}
               </div>
+              <div>
+                <a class="btn btn-outline-primary btn-canvas mx-4" href={{multi_export_link}} role="button" download="communities.geojson">Export GeoJSON</a>
+              </div>              
               <div id="headingOne">
                 <button class="viz-dropdown btn btn-light font-weight-light" type="button" data-toggle="collapse" data-target="#collapseOne"
                 aria-expanded="false" aria-controls="collapseOne" style="width: 100%">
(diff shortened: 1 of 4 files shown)


jessiejofielding:
> This is awesome, Jessie!!
> 
> I'm working on reviewing this now. One thing I thought I should share before a full review is that I'm getting an error on geojsonlint.com (shown below). This is after exporting from a community drive map page. Are you able to reproduce this error on your computer?
> 
> Screen Shot 2020-07-21 at 11 18 55 AM

Let me look into this right now - I also am pushing a quick clean up to remove some commented out code I shouldn't have left in :)

lauren-johnston:
One more quick comment - I was thinking that maybe the export GeoJSON button

> > This is awesome, Jessie!!
> > I'm working on reviewing this now. One thing I thought I should share before a full review is that I'm getting an error on geojsonlint.com (shown below). This is after exporting from a community drive map page. Are you able to reproduce this error on your computer?
> > Screen Shot 2020-07-21 at 11 18 55 AM
> 
> Let me look into this right now - I also am pushing a quick clean up to remove some commented out code I shouldn't have left in :)

Thanks - Also quick comment - maybe the "export as geoJSON" button could be moved up the page to below the campaign title? Up to you. Just thinking it may help people find it more easily, esp if there are lots of communities 

Screen Shot 2020-07-21 at 11 23 21 AM

jessiejofielding:
> One more quick comment - I was thinking that maybe the export GeoJSON button
> 
> > > This is awesome, Jessie!!
> > > I'm working on reviewing this now. One thing I thought I should share before a full review is that I'm getting an error on geojsonlint.com (shown below). This is after exporting from a community drive map page. Are you able to reproduce this error on your computer?
> > > Screen Shot 2020-07-21 at 11 18 55 AM
> > 
> > 
> > Let me look into this right now - I also am pushing a quick clean up to remove some commented out code I shouldn't have left in :)
> 
> Thanks - Also quick comment - maybe the "export as geoJSON" button could be moved up the page to below the campaign title? Up to you. Just thinking it may help people find it more easily, esp if there are lots of communities
> 
> Screen Shot 2020-07-21 at 11 23 21 AM

Good call, I hadn't thought about how it would get lost with lots of communities. 

I've also fixed the geojson issue - this exports a community on its own from its own map as a Feature and not as a FeatureCollection, because I don't see why it needs to be a FeatureCollection, but if it does, I can make some adjustments so that it's still a FeatureCollection and so that it works for the multi-community export. I did this only because I thought it would make for easier to manipulate geoJSONs for users who export their communities, but I'm not an expert on what people want from that... Let me know which you think is best and then I'll push these changes :)

jessiejofielding:
Pushed the change - the geoJSON files should now be valid, but to test, also go to "My Communities" and export the geoJSON of one of those communities just to make sure that change is in fact working

lauren-johnston:
Great! Playing around with this now. One thing that cropped up is this issue that we had before with the right hand rule. @k-y-le imported a package that helps fix this I believe

Screen Shot 2020-07-21 at 3 38 57 PM

jessiejofielding:
> Great! Playing around with this now. One thing that cropped up is this issue that we had before with the right hand rule. @k-y-le imported a package that helps fix this I believe
> 
> Screen Shot 2020-07-21 at 3 38 57 PM

Is this an issue on the single community geoJSON too?

jessiejofielding:
I can only duplicate the error when I export from the Drive map page when I have a community entry that I messed with in the database that has a drive but no organization - I think something is failing silently, but I was under the impression we weren't allowing entries to have only a drive or only an org?

jessiejofielding:
> I can only duplicate the error when I export from the Drive map page when I have a community entry that I messed with in the database that has a drive but no organization - I think something is failing silently, but I was under the impression we weren't allowing entries to have only a drive or only an org?

After further inspection: it's happening on drives that aren't created through a drive that then have a drive added in postico, regardless of whether an org is added, and I'm not sure it's always happening with those, and I'm not sure it's only happening on those but it seems to be. Kyle and I have been talking about it, and we'll deal with this tomorrow hopefully

jessiejofielding:
Okay, thanks to some help from Kyle, we've determined that this bug exists on main too, for some entries but not others, following no particular pattern (I suppose I was just looking for patterns). Kyle suggested two possibilities: geojson_rewind is not quite doing what it should be, or geojsonlint isn't. On the second point, I took some files that were supposedly not following the right-hand rule, and I had no trouble opening them in kepler.gl (through VSCode). Either way, since this bug exists on main, should we merge this PR (assuming that's the only issue left) and open a new one for the bug?

lauren-johnston:
> Okay, thanks to some help from Kyle, we've determined that this bug exists on main too, for some entries but not others, following no particular pattern (I suppose I was just looking for patterns). Kyle suggested two possibilities: geojson_rewind is not quite doing what it should be, or geojsonlint isn't. On the second point, I took some files that were supposedly not following the right-hand rule, and I had no trouble opening them in kepler.gl (through VSCode). Either way, since this bug exists on main, should we merge this PR (assuming that's the only issue left) and open a new one for the bug?

Yeah feel free to merge! I think this is the right call because there's at least a few major GIS platforms still allow the old geoJSON format.

PR #308, opened 2020-07-22
Add info to partner page, auto email when org is created

lauren-johnston:
Key Changes:
- Simplified org creation flow by remove request access. Now it's get started (takes you to org creation form) and request demo (for those who aren't sure yet). 
- Automatically emails team@representable.org when a user has created an org. 
- Includes user details in email (see screenshots) for easy follow up
- Updated info on partner page to reflect community mapping drives

To test:
- I tested this by checking the email in the console after creating an org at: http://127.0.0.1:8000/dashboard/partners/create/ 
- I also tried moving email settings from prod.py to base.py to make sure the email would actually send and it did for me.
- Play around with button behavior, review content at: http://127.0.0.1:8000/partners/welcome/

Screenshots:
Screen Shot 2020-07-22 at 12 13 53 PM
Screen Shot 2020-07-22 at 11 50 28 AM

### DIFF
--- a/main/templates/main/partners/get_started.html
+++ b/main/templates/main/partners/get_started.html
@@ -7,9 +7,10 @@
     <div class="col-lg-6 text-center my-5 order-md-1">
       <h1 class="font-weight-light"> Partner With Us</h1>
       <p class="font-weight-light">
-        Welcome to Representable! Request early access to create an organization below.
+        Welcome to Representable! Get started by creating an organization or requesting a demo below.
       </p>
-      <a class="btn btn-primary" href="mailto:team@representable.org?subject=Early%20Access&body=Dear%20Representable%20Team%2C%0D%0A%0D%0AI'd%20like%20to%20request%20early%20access%20to%20organization%20features%20on%20behalf%20of%20______." role="button">Request Early Access</a>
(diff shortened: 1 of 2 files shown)


ETedward:
This is awesome!! Email testing worked for me as well!

One small bug, that could also be addressed separately in another issue if you want; when creating a new organization the same one can be created multiple times; would be good to check if a user already created an organization, to avoid duplicates.

okusio:
This looks so awesome Lauren. Should we take out the moderate submissions part btw?

lauren-johnston:
> This is awesome!! Email testing worked for me as well!
> 
> One small bug, that could also be addressed separately in another issue if you want; when creating a new organization the same one can be created multiple times; would be good to check if a user already created an organization, to avoid duplicates.

Thanks for testing it out! Hmm - this is an interesting idea to limit the number of orgs to one per user. For flexibility, we haven't limited it although this is something we could consider in future.

lauren-johnston:
> This looks so awesome Lauren. Should we take out the moderate submissions part btw?

Ahh true!

lauren-johnston:
@okusio updated to remove moderation mention!

Screen Shot 2020-07-23 at 4 32 10 PM

PR #309, opened 2020-07-23
select tool

k-y-le:
**changes**
- removed mapbox draw tool (user polygon)
- implemented hover + click to highlight select tool
- change selection area with input control (1-100)
- 'Toggler Eraser' control un-highlights selected areas
- 'Clear Selection' control clears the selection
- made it so that user_polygon field is no longer required in db, and census_blocks_polygon_array is
- other changes to flow, including: wording of units above the map, removing the success message from displaying upon drawing a map
- this works for the current six states we have 'available' : mi, va, pa, nj, tx, il

**to test**
- go to entry page
- try to draw a district, play around with the tools
- save entry & check

**still to do (future PR)**
- validate contiguity of polygons (currently, can submit any selection (of any size as well) without issue, i wrote some comments on how to possibly move forward with turf.js#intersect but that's tbd)
- error messages + reloading the map on incomplete entry behaves a bit oddly currently

**screenshots**
image


lauren-johnston:
this is awesome!! taking a look now.

Btw, I can see that on Travis CI this test didn't pass. Definitely not a priority because we need to fix the form frontend/backend integration first, but for future reference: 

```
======================================================================
FAIL: test_invalid_polygon (main.test_entry.EntryTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/build/Representable/representable/main/test_entry.py", line 234, in test_invalid_polygon
    self.assertFalse(CommunityForm.is_valid(entry_form))
AssertionError: True is not false
```

k-y-le:
Thanks for the feedback! I'll update the test, since that test is for a user polygon with kinks, but we no longer test for kinks because we no longer gather the user polygon

> * Zooms in too far initially (depending on zipcode). This makes it unclear what's being highlighted.

I think this is mostly how the geocoder/zipcode works. Like certain zipcodes (08544 a good example) cover a small area, while others are much bigger. I can add a maxZoom but I think that may be unintuitive for people who are searching for their home address, or something else that zooms in close. so i'm not sure if there's a fix here; any ideas?

> * place names are hard to make out at times.
agreed, and i'm looking into a way to make them more apparent. i'll make the blockgroup lines more transparent, but in general there will have to be a tradeoff -- hopefully something in mapbox styles will help

> * "toggle eraser" could be changed to "eraser". And if possible, we want to show the user that they are currently drawing by including a draw button. this would make it similar to other tools like photoshop etc

I get what you mean, and this should be an easy addition. I'll add it!

k-y-le:
> agreed, and i'm looking into a way to make them more apparent. i'll make the blockgroup lines more transparent, but in general there will have to be a tradeoff -- hopefully something in mapbox styles will help

here's what it looks like if i place the census layers below the mapbox streets layers -- only issue is that some of the census lines are covered up by roads/other features.

image

also side note: i've decided to comment out the test in question, with the condition that it should be added as an entry test once we implement testing contiguity/area of census block polygons

ETedward:
This is so amazing!! Aside from the above comments, everything works great for me! Step by step is also now updated to reflect the draw and eraser changes.

PR #312, opened 2020-07-27
Email Verification in Organization Flow

ETedward:
**Changes**:

1) When an organization is created with a not verified account: Resends verification email link, displays modified message  "please verify account, email verification is necessary for certain features ..."

2) When a not verified account attempts to create a community drive: Does not let the account create a drive, displays message "email verification is necessary to access this feature ...", Resends a verification link.

**To test**

Make an organization with a not verified email account.
Check to see verification link is resent.
Check to see that a modified message is displayed.

Try creating a community drive with a not verified email account
Check to see verification link is resent.
Check to see that the following modified page is displayed instead of the Create a Drive form.

Screen Shot 2020-07-27 at 3 20 56 PM
11ea-91a5-8746cd7cd626.png">

### DIFF
--- a/main/fixtures/single_org.json
+++ b/main/fixtures/single_org.json
@@ -78,5 +78,15 @@
       "created_at": "2020-06-26T00:36:52.834Z",
       "is_active": true
     }
+  },
+  {
+    "model": "account.emailaddress",
+    "pk": 1,
+    "fields": {
+      "email": "test@me.com",
+      "verified": true,
+      "primary": true,
+      "user_id": 1
+    }
   }
 ]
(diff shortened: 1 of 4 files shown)


k-y-le:
Yep, when I ran `python manage.py test` I got an error for test create organization -- seems like the test fails because it tries to create an organization using an unverified acct (no Email associated w user)

Other than the tests it looks good to me!

ETedward:
Thanks great catch!

The error seems to occur because the setup of test_organization.py never has an email attribute associated with the account (only getting a username attribute). Would anyone know how to add an email attribute, or add an email to "test_user"?

lauren-johnston:
Test_organization uses a fixture(single_org.json). It’s loading the user,
including the email, from the fixture file when it calls the get_or_create
function. However, whether or not that user is verified is not considered
in my tests so that may be the problem?

PR #313, opened 2020-07-28
Delete orgs

### LINKED ISSUE #161: Org admins need the ability to delete organizations.
Is your feature request related to a problem? Please describe.
No way to delete organizations.

Describe the solution you'd like
There should be a clear way to delete organizations that warns the user they are deleting the campaign.

jessiejofielding:
**Changes:**

- update models so that deleting org doesn't delete all communities submitted to it
- create view for deleting organization
- create URL for deleting organization
- add button for deleting organization and pop-up modal for confirming the action

**To test:**
- migrate
- run server
- create test org to delete
- create drive within that org
- submit community to that drive
- go to the dashboard -> manage the org to delete -> press red trash icon
- confirm that close and cancel work
- delete permanently 
- check that community still exists under "my communities" with drive=NULL and organization=NULL
- check that the org and the drive are both deleted from the db 

Closes #161


lauren-johnston:
Awesome! Testing now. Love the confirmation message when you try to delete - will definitely prevent accidental deletion.

Two quick comments:
- May be worth adding a unit test or two to verify this works (I think this is probably optional for now given the amount of work you have on your plate, but it would make it easier to test features like this)
- small frontend issue here with the alignment of the trash can:
Screen Shot 2020-07-27 at 6 15 56 PM

jessiejofielding:
I meant to fix that - how do we feel about it here?

Also, I'll add some unit testing to my plans for next week, which should be a little quieter once I finish up the flagging MVP :) 

image

lauren-johnston:
> I meant to fix that - how do we feel about it here?
> 
> Also, I'll add some unit testing to my plans for next week, which should be a little quieter once I finish up the flagging MVP :)
> 
> image

Sounds good!

PR #318, opened 2020-07-30
check selection contiguity

k-y-le:
** changes **
- when users try to select an area which is disjoint from their current selection, displays a warning
- of note: this is not a perfect test. there are a decent amount of edge cases where it allows users to make a selection which is not connected to their current selection, but is pretty close. there are no false negatives (e.g. doesn't allow users to make selections which are contiguous), so this shouldn't impact the user flow much.
- this is because it uses bounding boxes, rather than the actual polygons. this greatly reduces the strain on the browser that checking the overlap of census blocks would cause

** to test **
- go to entry page
- make a selection. confirm that this does not produce an error message
- make a contiguous selection. confirm that this also does not produce an error.
- make a non-contiguous selection (somewhere disjoint from current selection). confirm that this is not allowed, and that a warning box pops up on the map.

** screenshots **
image


k-y-le:
Closing this PR as the changes are already implemented in #328

PR #319, opened 2020-07-30
Flagging mvp

### LINKED ISSUE #227: Organizations need to easily know which submissions are relevant to their state by automatically filtering submissions.
Problem:
Given a large volume of submissions, organizations will have a hard time sorting through submissions that are outside of their state.

Question:
How do we flag/hide submissions that are out-of-state? and how do we validate - by address, zipcode, or the boundaries of the community itself?

### LINKED ISSUE #229: Organizations do not want the power to delete submissions submitted to them (based MI feedback).
Issue:
Organizations do not want the power to delete submissions submitted to them (based MI feedback).

Possible solutions: 
We should consider this feedback when Django organization permissions are reviewed. See #224

### LINKED ISSUE #283: Add flag feature (model + frontend)
Users and organizations need a way to flag/report submissions as spam, abuse, etc.

- Create basic report model with community id, resolved = True or False
- Set all communities to admin approved by default
- Create a "flag community" button under each community on map pages
- Once a community is flagged, we would take a look from the Django Admin and un-approve submissions deemed abusive
- Remove review page and move to flagging communities from map [Kyle suggestion]
- Automatic address verification (removes the need for orgs to review submissions)

Previously had organization review pages

**Additional context**

jessiejofielding:
**Changes:**
- add report model
- add report button on partner maps (organization and drive) viewable to everyone
- add report pop up to confirm report, with required email input if user is not logged in
- add thank you pop up to confirm that the report was reported
- add report url
- add report view
- add automatic unapproval of a community if the user who made the report is an org admin
- add report admin view
- allow site admins to filter by reports resolved
- add report admin links to the community (only in the list)
- add report admin list action "unapprove and resolve", which for each report selected unapproves the community linked to and resolves the report
- remove review pages 
- add tests for reports

**To test:**
- migrate
- python manage.py test
- runserver
- go to org or drive map
- report a community, check that the frontend does what you'd expect
- go to admin page, under main, click on report
- see if the admin report list is satisfactory 

Closes #283 and closes #229


jessiejofielding:
Give me a few minutes to fix my tests :)

lauren-johnston:
Checking this out now!

lauren-johnston:
Looking good so far - quick comment, I think it may be  better to simply say "report" rather than the icon. I think this will help people understand how to report a submission. This could be a link in red potentially instead of a button to make it compact. Screen Shot 2020-07-31 at 11 00 30 AM

jessiejofielding:
> This looks really good! One thing that I wasn't able to verify was whether or not org admins can unapprove submissions automatically when they report them. It seems that there is no way to see if a community is approved or not as an org leader. I think there should be 1) a message that lets org admins know that communities they report will be hidden. 2) Org admins should be shown all submissions with clear indication of whether they're approved or not. Non admins and everyone else should only see approved submissions. Thoughts?

I absolutely agree - I'll add the message and fix the button real quick! Do you think a simple "approved"/"unapproved" tag similar to the drives (if you're on the organization map) would do the trick? Is that the language we want to use? Do we just want a "flagged" or "hidden" tag that shows up if the community is unapproved? Also, in this system we want to create automatically approved submissions, right, and only unapprove if necessary? I think that we're not currently doing that, so I should fix it if that's what we want...

lauren-johnston:
> > This looks really good! One thing that I wasn't able to verify was whether or not org admins can unapprove submissions automatically when they report them. It seems that there is no way to see if a community is approved or not as an org leader. I think there should be 1) a message that lets org admins know that communities they report will be hidden. 2) Org admins should be shown all submissions with clear indication of whether they're approved or not. Non admins and everyone else should only see approved submissions. Thoughts?
> 
> I absolutely agree - I'll add the message and fix the button real quick! Do you think a simple "approved"/"unapproved" tag similar to the drives (if you're on the organization map) would do the trick? Is that the language we want to use? Do we just want a "flagged" or "hidden" tag that shows up if the community is unapproved? Also, in this system we want to create automatically approved submissions, right, and only unapprove if necessary? I think that we're not currently doing that, so I should fix it if that's what we want...

I think showing a "hidden" tag sounds good. However, I can't see the tag on either the drive or org maps. Also yeah I think it's a good idea to reverse this "Also, in this system we want to create automatically approved submissions, right, and only unapprove if necessary" as you mentioned.

jessiejofielding:
Latest push should make all the necessary changes :)

PR #328, opened 2020-08-04
entry page mapping and validation streamlining

k-y-le:
**changes**
- adaptation to mapping code to decrease load on client side throughout the drawing process (creates polygon after pressing submit, instead of during the drawing process)
- refreshing the page or a validation failure now maintains the drawn community (don't have to start over)
- final test for contiguity, after submitting
  - we now have two tests for contiguity -- first, when selecting areas, a test to ensure that the selected units border one another. secondly, for anything not caught by this first test (ex: drawing a community then erasing part so that you have two separate areas) will be caught as an error when submitting
  - note that this contiguity test does not impact communities with a hole in the middle -- this case is noted in the **next steps** section lower down
- test added for sizing of the community :: if a community has over 800 blockgroups, form validation failed
- fixed form validation :: alerts for the drawn map display and scroll to, alerts for various form fields now highlight only the unfilled area, and scroll to the top
- added a "saving community..." popup upon pressing save to account for loading times to load communities

**to test**
- go to entry page
- fill out form normally, check that it succeeds
- fill out map with community that has >800 blockgroups, check that it raises an error
- fill out map with community that has two or more distinct parts, check that it raises an error
- fill out form missing different fields, check that it raises an error and only highlights the missing field

**screenshots**
- community size check
image
- non-contiguous check
image

**next steps**
- per discussion on `community_entry` model changes, this code sets up nicely to allow for saving only the external points of a polygon (fixes holes in community, using similar code to that which displays polygons on map.js), rather than the WKT polygon right now
- the `polyFilter` array in `createCommPoly()` method contains an array of GEOIDs for each blockgroup in the community, which can be easily loaded into the db upon saving + model changes


jessiejofielding:
Quick question - when I get a validation error that isn't about the map, my drawn community does persist but the map disappears so I have to search for something again - do we mind this?

Other than that, it looks good! I can't select block groups that aren't touching my current selection (is this intended? it tells me why the selection isn't working, so I have no qualms with it, I just wasn't sure based on your wording if I should be able to), but when I erase a gap in the middle, I get the correct validation error :)

PR #330, opened 2020-08-04
Simplify toggle behavior 

lauren-johnston:
Key changes: 
-Instead of auto toggling draw/erase on every click, this PR only toggles if a new button is selected

To test:
- Visit entry page
- Inspect js for errors


lauren-johnston:
> looks good - only issue is that if i select the eraser tool and then go to the draw tool, it doesn't deselect the eraser :
> image

Thanks will fix this!

lauren-johnston:
Fixed! Wasn't working because backgroundColor was spelled "backgrundColor"

PR #331, opened 2020-08-04
Initial Map for Entry_State_View

ETedward:
**Changes**
For entry pages associated with a drive, or associated with a state - have the map immediately display zoomed into the center of that state. 
Also fixed small bugs and wordings in step by step guide (ie. removed "radius bar" terminology)

**To Replicate**
1) From entry page, click on a state that has a state page (ie. Michigan). Click "draw my community" or join one of the drives. Alternatively, make your own org and drive, and navigate to the associated entry page.
2) Check that the map on the entry page is displayed initially zoomed to the associated state.
3) Check that census blocks are initially displayed for the associated state.
4) Play around with select tool to make sure things work as usual. Also click step by step guide and check that the terminology and steps make sense.

### DIFF
--- a/main/static/main/js/geo.js
+++ b/main/static/main/js/geo.js
@@ -738,6 +738,29 @@ map.on("style.load", function () {
     newHighlightLayer(states[i]);
   }
 
+  // Initialize Map for State Pages
+  if (state_abbr !== "") {
+    showMap();
+    map.flyTo({
+      center: statesLngLat[state_abbr],
+      zoom: 6,
+      essential: true, // this animation is considered essential with respect to prefers-reduced-motion
+    });
+
+    state = state_abbr.toLowerCase();
+    $("#shepherd-btn").removeClass("d-none");
+    map.setLayoutProperty(state + "-census-lines", "visibility", "visible");
+
+    // Save state to session storage
+    sessionStorage.setItem("state_name", state);
+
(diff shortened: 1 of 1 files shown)


k-y-le:
> It looks like upon first entering the state page, the hover functionality isn't working. It starts working after using the geocoder. I'm looking into it further!

I pushed a fix for this! looks good.

The one thing i ran into is that when manually entering in a url (e.g. http://127.0.0.1:8000/entry/IL/ ), it doesn't load the map. on my machine, the only states where the map is displayed immediately are Michigan and Texas. Is this as intended?

ETedward:
Thanks for the fix Kyle!
Yes, this is intended! only pages that have a state page created will have a url link and map displayed.

ETedward:
Sadly ran into a merge conflict. Merged the entry_state_views branch but now there's a bug in main with all the units initially selected for the associated state :(

lauren-johnston:
> Sadly ran into a merge conflict. Merged the entry_state_views branch but now there's a bug in main with all the units initially selected for the associated state :(

I just reverted the merge to main so we can fix this!

ETedward:
Thanks! 
I don't know if it's the best fix, but adding the following code into the bottom of the part of // Initializing Map for State Pages seems to fix the issue.

sessionStorage.setItem("bgFilter", "[]");
sessionStorage.setItem("selectBbox", "[]");

PR #338, opened 2020-08-06
resolving initial state map merge conflict

ETedward:
**Changes**
Bug: When navigating to an initial state all the units of that state are selected
Fixed the bug (clears selectbBox)

**To Replicate**
Navigate to an entry page associated with a state
Check to see that the units are not all initially selected
Play around with the select tool to see that things work as usual;.

### DIFF
--- a/main/static/main/js/geo.js
+++ b/main/static/main/js/geo.js
@@ -971,58 +971,6 @@ map.on("style.load", function () {
     // Save state to session storage
     sessionStorage.setItem("state_name", state);
 
-    // When the user moves their mouse over the census shading layer, we'll update the
-    // feature state for the feature under the mouse.
-    var bgID = null;
-    var features = [];
-    var stateCensus = state + "-census-shading";
-    // if touch screen, disable.
-    if (!is_touch_device()) {
-      map.on("mousemove", stateCensus, function (e) {
-        if (e.features.length > 0) {
-          // create a constantly updated list of the features which have been highlighted in foreach loop
-          // before highlighting, go thru that list, and deselect all
-          var bbox = [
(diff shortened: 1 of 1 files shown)


lauren-johnston:
Hey this is awesome! Would you mind resolving the conflicts in geo.js before I take a look (just to make sure that I can fully test it and make sure it works)

ETedward:
Sorry, yes conflicts resolved!

ETedward:
Hmm at the moment I'm still able to see the grey hover, just to make sure this greyed area is what we are referring to?
Screen Shot 2020-08-10 at 1 39 54 PM

lauren-johnston:
> Hmm at the moment I'm still able to see the grey hover, just to make sure this greyed area is what we are referring to?
> Screen Shot 2020-08-10 at 1 39 54 PM

I can see the hover on localhost/entry but not localhost/entry/mi

k-y-le:
just took a look at this and it seems that there was a mismatch between the other branch and this one? the problem with the hover is because the mapbox code for on mousemove is called only upon geocoder result. i can push a quick fix!

k-y-le:
also merged with main, but i do have one note -- if i go to /entry/MI/ and then search in the geocoder for Princeton, NJ -- then it still displays the lines for block groups in new jersey.. I think the way it should behave is that if you're on a link like entry/MI/ , then only the lines for michigan will ever load, even if you search for a location outside of the state

ETedward:
Thanks so much for the fixes and merges!

Yes I've implemented the proposed change where for /entry/MI/ only census blocks lines in Michigan will be displayed even when the user searches for locations elsewhere such as New Jersey or Texas.
And where on /entry/ census block lines are displayed per usual based on the search.

PR #340, opened 2020-08-09
entry page undo button

k-y-le:
**changes**
- added an undo button on the entry page with the following behavior
  - after an action which draws, erases, or clears the map, allow to undo the previous action
  - if there is no action to undo, raise a warning
  - undo only applies for actions which impact the drawn community (ex. using the eraser but not deselecting any area doesn't apply)
- trying to clear an empty map raises a warning
- updates selection area upon using eraser tool or undo tool to fix edge cases for contiguity test + submission
- fix bugs on submission page

**to test**
- go to entry page
- search to begin drawing community
- try clearing an empty map. test that it raises a warning
- try undo-ing a non-action. test that it raises a warning
- test out drawing and subsequently undoing actions
- submit & confirm working submission action

**screenshots**
image


ETedward:
This is amazing!! worked great for me!

PR #341, opened 2020-08-09
Added tooltips + changed mapbox style

k-y-le:
**changes**
- per request from IL partners, added a tooltip on state page which displays more information on community mapping drives
- changed "units" link on entry page to also include a tooltip with a brief description of what census block groups are / why we use them
- added a custom mapbox style to the entry page, which displays the maximum amount of landmarks/roads/place names without clutter

**screenshots**
image
image


ETedward:
Looks good to me!! just to make sure none of the census and other keys removed during clean up are still used?

Screen Shot 2020-08-11 at 2 09 49 PM

k-y-le:
> Looks good to me!! just to make sure none of the census and other keys removed during clean up are still used?
> 
> Screen Shot 2020-08-11 at 2 09 49 PM

nope! they're all deprecated now

PR #351, opened 2020-08-11
test community size while drawing

k-y-le:
**changes**
- small change adds a size test while drawing (rather than on submit) to warn users if they are selecting a community that is too large

**to test**
- go to entry page
- try to select community over 800 block groups in size
- check warning message shows, no js error, etc

**screenshot**
image


jessiejofielding:
It looks good, and no js errors... however, hovering over the blockgroups doesn't turn them grey anymore?

jessiejofielding:
> It looks good, and no js errors... however, hovering over the blockgroups doesn't turn them grey anymore? I presume because of one of the other PRs that just got merged with main?

Update: hovering is working on main

k-y-le:
> > It looks good, and no js errors... however, hovering over the blockgroups doesn't turn them grey anymore? I presume because of one of the other PRs that just got merged with main?
> 
> Update: hovering is working on main

I was able to replicate on main (happens when you search for a different state than previous) -- is this what you were experiencing?

this PR shouldn't impact any of that, so i think it would make sense to merge this and open an issue for the hover bug

jessiejofielding:
> > > It looks good, and no js errors... however, hovering over the blockgroups doesn't turn them grey anymore? I presume because of one of the other PRs that just got merged with main?
> 
> > 
> 
> > Update: hovering is working on main
> 
> 
> 
> I was able to replicate on main (happens when you search for a different state than previous) -- is this what you were experiencing?
> 
> 
> 
> this PR shouldn't impact any of that, so i think it would make sense to merge this and open an issue for the hover bug

I had the problem on the first state I searched when I was at just /entry/, but is probably related and since this PR isn't related, I'm going to go ahead and approve it

PR #355, opened 2020-08-12
store only outer coordinates in community entry polygon

### LINKED ISSUE #353: Implement community entry model change -> store only outer coordinates

lauren-johnston:
Key changes:
- Runs the function below on polygons in multipolygon
- resolves #353 
```
function getPolygonOutline(polygon) {
  // check how deeply nested the outer ring of the unioned polygon is
  newPolygon = [];
  for (obj in polygon) {
    outline = [];

    // set the coordinates of the outer ring to final
    if (polygon[obj][0][0].length > 2) {
      outline = [polygon[obj][0][0]];
    } else if (polygon[obj][0].length > 2) {
      outline = [polygon[obj][0]];
    } else {
      outline = polygon[obj];
    }
    newPolygon.push(outline);
  }
  return newPolygon;
}
```

To test:
- Create a few polygons and save
- inspect JS console for errors


k-y-le:
no errors but it doesn't seem to be altering the polygon's coordinates in this case; the function getPolygonOutline() is returning the same coordinates as are passed to it, at least for this community with a hole : 
image

lauren-johnston:
> no errors but it doesn't seem to be altering the polygon's coordinates in this case; the function getPolygonOutline() is returning the same coordinates as are passed to it, at least for this community with a hole :
> image

Thanks was wondering about that - will take a look again

PR #373, opened 2020-08-20
Terms and privacy update

lauren-johnston:
Key Changes:
- Purchased license via getterms to use official adaptation of terms, privacy policy
- Removed getterms mention given commercial license

To test:
- N/A - visit terms and privacy to preview


k-y-le (approved):
🐕 💸

PR #380, opened 2020-08-22
Translations and Setup

ETedward:
**Updates**
Went through and internationalized entry page code. Natalie will fork from main and add her translations by Monday; agreed during the meeting to push a base version of translations to main before then.
Introduced {%blocktrans%} tags instead of {% trans"" %} which allows the preservation of <bold> <br> and hyperlinks within the translations.

**Previous Changes**
Added Spanish language translations for navigation bar and home page.
Added language button in footer to switch languages.

**How to replicate**
Run Python manage.py compilemessages -l "es"
Then run python manage.py runserver.
Navigate to home page. Switch to en espanol in the footer. Everything, except our slogan, should translate.
Navigate to entry page. And click en espanol. Nothing should change (translations are left empty in the .po file). 
Play around with entry page ie. see examples, submit a map, to make sure there are no weird symbols, "{%" or django translation tags that appear anywhere.

**Upcoming Features**
Key takeaway during language access meeting with Change Illinois: a) get a version of Spanish translations up as soon as possible; b) redesign switch language button to be in header and work with a native translator to improve translations.

1) After Natalie adds entry page translations we will work with a native speaker to go over translations and also get a translation of our slogan in Spanish.

2) We will redesign the switch language button to appear in the header. This will come in a future PR and will also depend on Lauren's work redesigning our header.


lauren-johnston:
Looks great! My only comment is that I noticed a few places where spanish translation isn't happening:

Screen Shot 2020-08-24 at 12 40 44 PM
Screen Shot 2020-08-24 at 12 40 53 PM
Screen Shot 2020-08-24 at 12 42 08 PM
Screen Shot 2020-08-24 at 12 42 58 PM
Screen Shot 2020-08-24 at 12 42 51 PM

PR #381, opened 2020-08-23
Added in Spanish translations for index and entry

natastro:
compared Google Doc with .po file and updated translations. E-mailed @lauren-johnston and @ETedward with questions regarding other translations on the .po file

Looks like a lot of index was filled in but I went in and compared the two. I'm happy to go in and keep translating anything with `msgstr` 😄 Have a great weekend!

### DIFF
--- a/representable/<representable>locale/es/LC_MESSAGES/django.po
+++ b/representable/<representable>locale/es/LC_MESSAGES/django.po
@@ -2163,7 +2163,7 @@ msgstr ""
 
 #: main/templates/main/entry.html:64
 msgid "Map your community"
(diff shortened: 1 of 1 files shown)


ETedward:
This is awesome!!

ETedward:
Overall fantastic stuff especially with new lines and html formatting in the .po file. 
Just merged the changes and updated the entry page code to get the new translations working!
(although now there are a couple new untranslated sections in the .po files)

natastro:
Hi @ETedward ! Happy to take care of those as well today. I only took care of entry ones (index was filled out but I double checked them to make sure they weren't missing anything). Just let me know 😄

PR #382, opened 2020-08-23
More Spanish translations!

natastro:
changed out some French translations, filled in what was remaining for 'entry.html'

Thanks again @ETedward !


ETedward:
Looks great! thanks @natastro

PR #383, opened 2020-08-23
entry page state views only

### LINKED ISSUE #361: State entry page user flows
Currently, if a state does not have a state page, users are redirected to /entry/, instead of /entry/{abbrev}

- All attempts to draw a community go to a state specific entry page (/entry/ deprecated)
- Header button "Add community" redirects to state selection area

k-y-le:
users can now only draw a community when going through the flow of selecting a state. closes #361 

**changes**
- going to /entry/ now redirects to the state selection on home page
- clicking on the header "Add Community" button now redirected to state selection on home page
- removed listening for geocoder response on entry page
- geo.js code now uses mapbox more efficiently
- cleaned up some deprecated css

**to test**
- click on "add community" header button
- try to go to "/entry/" manually
- go to entry page, check that it loads correctly when using geocoder


k-y-le:
> This is awesome! I haven't gotten JS errors so far but before I review further, I noticed an issue with the geocoder. First, I selected missouri on the map. Then I went to entry/sc directly and weirdly enough it still went to missouri even though I asked for south carolina. It seemed like the map zoomed to SC first and then actually moved again to MO.
> 
> Screen of when I go to entry/sc:
> Screen Shot 2020-08-24 at 12 49 23 PM

good catch -- it seems this happens when one draws on the map, since it zooms to that draw area, even if the state has changed. i pushed a fix!

lauren-johnston:
Amazing will take another look!

lauren-johnston:
I think the geocoder issues is fixed! awesome. 

Seeing the error below in the travis CI now, think it may be in the view code (screenshotted below)?

Screen Shot 2020-08-24 at 8 33 16 PM
Screen Shot 2020-08-24 at 8 34 41 PM

k-y-le:
pushed an update to tests which tests /entry/mi/ instead of /entry/ 👾

PR #384, opened 2020-08-25
display population data on map drawing

### LINKED ISSUE #357: Users want population data to display when drawing the map
Users have expressed interest in being able to see the population of their community while drawing.

After selecting an area of the map, users see a population number for the area they've selected.

Previously, we had population data which compared it to current district sizes. We no longer want this, but displaying just population is useful.

k-y-le:
closes #357 
**changes**
- users can now see the population of the map when drawing their community (queries census api)
- displays "Community population: ..." while loading

**to test**
- `vim ~/.bash_profile` -- insert `export CENSUS_API_KEY:"{key}"` with the census api key (reach out to me if you don't have the key - in bitwarden as well)
- `source ~/.bash_profile` to reload key
- run server, go to entry page
- check population appears when selecting an area (takes longer depending on size of community -- will work on efficiency at a later time)
- check no js errors

**screenshots**
image

**other considerations**
- currently, reloading the entry page after drawing a community means the community is erased, but the population number still shows. this is fixed in #383 , so shouldn't be an issue after merging


jessiejofielding (changes requested):
Again, not requested but it's been here a while: when I try to add a community, selection doesn't work and I get these errors

image

arorasomya:
tried testing locally, and population is not calculated:
image

k-y-le:
Reopening this with some changes!
- adds a population field to the CommunityEntry model for storing the community population
- displays said population on drive map and submission pages (only for communities drawn with population calculated)
**todo**
- calculate population retroactively for already drawn communities?
- think about the best way to display population data : probably in conjunction with other demographics in the future !!
**screenshots**
image
image

PR #393, opened 2020-09-03
Remove Drive links on state page

lauren-johnston:
Key Changes:
- Temporary redesign of state page while we reconsider flow, removes community drive links

To test:
- Visit any state page
- python create_state_fixtures.py on local or visit admin -> states -> editor for state of choice

Screenshot:
Screen Shot 2020-09-03 at 7 16 05 AM

### DIFF
--- a/main/templates/main/state.html
+++ b/main/templates/main/state.html
@@ -10,43 +10,26 @@
   });
 </script>
 <div class="container">
-  <div class="row row-hero my-5">
-    <div class="col-12">
+  <div class="row my-5">
+    <div class="col">
       <h1 class="font-weight-light text-center mb-3">Welcome to {{ state_obj.name }}
       </h1>
+    <div class="text-center">
+      <a class="btn btn-primary mb-3" href="{% url 'main:entry' %}{{state_obj.abbr|lower}}" role="button">Draw my Community</a>
+      <a class="btn btn-outline-primary mb-3" href="#more" role="button">Learn More</a>
     </div>
-  </div>
-
-  <div class="row">
-    <div class="col-sm-6">
+      <a name="more"></a>
       {% autoescape off %}
       {{ state_obj.content_news }}
       {% endautoescape %}
-    </div>
-    <div class="col-6 d-none d-sm-block">
-      {% with "img/states/"|add:state_obj.abbr|add:".svg" as svg %}
-      <div class="card card-signin mx-auto h-100 py-4">
-        <object type="image/svg+xml" class="state h-100" data="{% static svg %}"></object>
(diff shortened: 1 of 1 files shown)


lauren-johnston:
Note - the MI page above does not appear on 1 page (you need to scroll to see the button at the bottom). I wanted to include button at top for mobile users especially.

k-y-le (commented):
Looks good to me, I'm just a bit worried about the visuals -- this is a pretty daunting wall of text, and we made the decision to include the state svg & the accordion so that people are more likely to read & not skip over everything immediately -- what are your thoughts?

I definitely really like having the "draw my community" button at the very top, for emphasis!

lauren-johnston:
> Looks good to me, I'm just a bit worried about the visuals -- this is a pretty daunting wall of text, and we made the decision to include the state svg & the accordion so that people are more likely to read & not skip over everything immediately -- what are your thoughts?
> 
> I definitely really like having the "draw my community" button at the very top, for emphasis!

gotcha I was anticipating that people would skip the text altogether for the most part and just draw from the top. The state svg just makes the content much longer imo so I removed for clarity and also because it broke the layout depending on length of the redistricting news" section. thoughts? I don't have strong feelings about the accordion, could add it back

k-y-le:
> > Looks good to me, I'm just a bit worried about the visuals -- this is a pretty daunting wall of text, and we made the decision to include the state svg & the accordion so that people are more likely to read & not skip over everything immediately -- what are your thoughts?
> > I definitely really like having the "draw my community" button at the very top, for emphasis!
> 
> gotcha I was anticipating that people would skip the text altogether for the most part and just draw from the top. The state svg just makes the content much longer imo so I removed for clarity and also because it broke the layout depending on length of the redistricting news" section. thoughts? I don't have strong feelings about the accordion, could add it back

makes sense - let's loop in @okusio ? she may have thoughts that add clarification

lauren-johnston:
> > > Looks good to me, I'm just a bit worried about the visuals -- this is a pretty daunting wall of text, and we made the decision to include the state svg & the accordion so that people are more likely to read & not skip over everything immediately -- what are your thoughts?
> > > I definitely really like having the "draw my community" button at the very top, for emphasis!
> > 
> > 
> > gotcha I was anticipating that people would skip the text altogether for the most part and just draw from the top. The state svg just makes the content much longer imo so I removed for clarity and also because it broke the layout depending on length of the redistricting news" section. thoughts? I don't have strong feelings about the accordion, could add it back
> 
> makes sense - let's loop in @okusio ? she may have thoughts that add clarification

Sg!

PR #395, opened 2020-09-08
Remove contiguity check

### LINKED ISSUE #388: Contiguity failure fails on some valid side-by-side census block groups
**Describe the bug**
Cannot select a census block group that is clearly next to another one (Nevada). A user reported this happening in VA as well.

**To Reproduce**
1. Go to entry page for nv:
2. Click on census block groups shown in screenshot
3. See error

**Expected behavior**
Contiguity check only fails on communities that are not connected.

**Screenshots**
Screen Shot 2020-08-26 at 1 07 37 PM

**Desktop (please complete the following information):**
 - Browser: Chrome

k-y-le:
Closes #388 
**changes**
- removed contiguity check on entry page (per Fairfax, VA bug)
- model changes: census_blocks_polygon_array no longer required, and census_blocks_polygon is now a GeometryField (allows for both polygons and multipolygons)

**to test**
- python manage.py migrate
- draw a community that generates a polygon (most communities)
- draw a community that generates a multipolygon (ex: fairfax comm below)
- check no errors

**screenshot**
previously, this exact community in Fairfax, VA triggered the contiguity check (and beyond that, wouldn't save) because it generated a multipolygon when applying turf.union(), rather than a polygon. This is the only example we've found thus far in which this is the case (bug found by Dave from DRA)
image


jessiejofielding (approved):
I get no errors! This looks good, thanks for getting this sorted

PR #396, opened 2020-09-08
moved data layers beneath existing map layers

### LINKED ISSUE #391: Census block groups hidden below other layers
**Describe the bug**
See screenshot below.
Screen Shot 2020-09-01 at 4 51 39 PM

k-y-le:
Closes #391 
**changes**
- data layers moved beneath existing map layers, as per bug report 391

**to test**
- go to map page or submission page, check lines are visible above map layers
(to determine in the future -- do we want these lines thicker? we had them that way before, but if i'm being honest, it was really ugly)

**screenshot**
image

### DIFF
--- a/main/static/main/js/map.js
+++ b/main/static/main/js/map.js
@@ -59,7 +59,7 @@ function newSourceLayer(name, mbCode) {
 }
 
 // add a new mapbox boundaries source + layer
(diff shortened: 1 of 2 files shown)


jessiejofielding (approved):
I know I wasn't requested on this, but I saw it had been here for a while, so I tested it and it looks good to me!

PR #399, opened 2020-09-10
Update yarn.lock

jessiejofielding:
**Changes:**
- update serialize-javascript to latest version

**To test:**
- there shouldn't be any changes - we require the latest version in some files but not others, so this shouldn't break anything


k-y-le (approved):
🥦 👍

PR #400, opened 2020-09-11
remove neighborhoods data layer

k-y-le:
as per feedback from CA, removed the neighborhoods data layer for being confusing/incomplete

**to test**
- go to organization map / submission page
- check no neighborhoods data layer


jessiejofielding (approved):
Looks good!

PR #410, opened 2020-09-24
More Spanish Translations

ETedward:
**Changes**
Added more Spanish translations from Natalie!
Added update to title in About page for me.
Will be working with Natalie this and next week to hopefully get our About page also translated!

**To replicate**
Run Python manage.py compilemessages -l "es"
Then run python manage.py runserver.
Switch to en espanol in footer and navigate to entry page and home page.
Make sure no weird symbols pop up, "{%" or django translation tags.


jessiejofielding (approved):
Looking good!

PR #412, opened 2020-09-28
add founding team section to about page

k-y-le:
adding the names of the founders and a short blurb about how Rep started to the about page
image


arorasomya (approved):
Looks good!

PR #413, opened 2020-09-28
various accessibility updates

k-y-le:
**changes**
- added html alt-text, aria-labelledby, role, etc for accessibility
- added labels to form fields
- updated translation button in footer to fit in better


jessiejofielding (approved):
This looks good!

PR #418, opened 2020-10-05
drive page + signup user flow updates

k-y-le:
a couple changes made to make user flow more smooth
**changes**
- updated the drive page to contain more information about representable + redistricting info for the state the drive is within
- changed user flows to redirect to signup page rather than login initially
- adapted links to ensure that "?next=" part of url is maintained between signup and login pages
- cleaned up frontend of signup and login pages
- removed michigan specific views/urls code

**to test**
- go to drive url
- check no errors / typos
- log out
- return to drive page or click on state from landing page
- check that you are redirected to signup page
- check that switching between signup and login pages maintains "?next=" part of url

**screenshots**
drive page
image


jessiejofielding (approved):
This looks great - just to check, only the entry page redirects to sign up? The drive and state pages do not require login and so didn't redirect me to sign up, which I thought was still what we wanted? If I'm correct, then it's good to go!

k-y-le:
> This looks great - just to check, only the entry page redirects to sign up? The drive and state pages do not require login and so didn't redirect me to sign up, which I thought was still what we wanted? If I'm correct, then it's good to go!

that's right - works as intended, i think. merging now

PR #437, opened 2020-10-10
About Page Translations

ETedward:
**Changes:** 
About Page Spanish Translations
**To Replicate:** 
Run "Python manage.py compilemessages -l "es""
Run "Python manage.py runserver"
Navigate to About page and enjoy the translations!


k-y-le (commented):
Looks good and works well - one note, though, is that the current about page (representable.org/about) is updated a biit from this one - section at the bottom with the founding team

Issue #423, opened 2020-10-06
Rethink "selection size" tool


k-y-le:
**Is your feature request related to a problem? Please describe.**
Currently, the selection size control on the map part of the entry page can be a bit confusing

**Describe the solution you'd like**
Selection size aspect of the map should use less confusing terminology and have UX elements that make it as self explanatory as possible

**details**
- remove number alongside bar
- make select bar go from small circle to larger one, to indicate a larger radius
- [future] add visual radius around cursor on map to display selection radius

k-y-le:
Closed, added via recent PRs #438 #444

PR #438, opened 2020-10-13
Update selection tool UX

k-y-le:
as per discussion in team sync + #423 
**changes**
- removed number for radius of selection tool
- changed to "selection tool size" from "selection size"
- slider thumb changes in size based upon overall selection size

**to test**
- go to entry page
- try out the selection tool, confirm no errors

**screenshots**
small
image
bigger
image

note: it's ever-so-slightly off center vertically... this is because css is really annoying about aligning this item. i think it's probably ok for now! but, we should keep an eye on..


k-y-le:
updated: fixed alignment issues - should appear solidly aligned vertically now

jessiejofielding (approved):
This is looking great, Kyle!!

PR #439, opened 2020-10-14
Dev reqs

jessiejofielding:
**Changes:**

- fix the fact that certain packages can't load on server
- create dev-reqs file for requirements that are only necessary for dev environments
- working from bug fix from last week so there's nothing to test


k-y-le (approved):
looks good!

PR #444, opened 2020-10-20
Simplified mapping controls

k-y-le:
**changes**
- changed initial mapping form to a basic mode wherein users can select + deselect individual census block groups, one at a time. Click to select. Click to deselect.
- added a dropdown settings button, which when pressed allows users to interact with the same mapping controls as before (selection size, eraser, etc)
- when closing the dropdown, keeps the same selection size/eraser/draw options, with the exception of the case where the selection size is 0, or minimum

**to test**
- go to entry page
- mess around with drawing a community
- check contiguity test (can you draw communities of geographically discrete block groups?)
- check that mapping behaves as expected when switching between the settings dropdown

**note**
- plan is to merge this into staging so that both sites can be used for user testing. Making this PR to make it open for review before/during said testing

### DIFF
--- a/main/static/main/css/style.css
+++ b/main/static/main/css/style.css
@@ -558,8 +558,8 @@ color scheme. See readme.md*/
 }
 input[type=range]::-webkit-slider-thumb {
   border-radius: 50%;
-  width: 17.5px;
-  height: 17.5px;
+  width: 15px;
+  height: 15px;
 }
 
 /* css for the line on the progress bar */
--- a/main/static/main/css/style.css
+++ b/main/static/main/css/style.css
@@ -556,6 +556,12 @@ color scheme. See readme.md*/
   /* font-family: Inter, system-ui, sans-serif; */
   /* font-size: 12px; */
 }
+input[type=range]::-webkit-slider-thumb {
+  border-radius: 50%;
+  margin-bottom: 6px;
+  width: 17.5px;
+  height: 17.5px;
+}
 
 /* css for the line on the progress bar */
 .bar-step {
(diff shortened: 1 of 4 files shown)


arorasomya:
Couldn't find bugs, but one question about the settings menu (and @okusio  might be able to talk about this better):but it was kinda confusing when the ability to select and deselect changed to select, but change to eraser to deselect. Might wanna think more about this? Would love to know your thoughts @okusio given ur user testing trials?

okusio:
Yes, we thought a bit about this. Users definitely found it easier to just click to select and click again to deselect. However, once you increase the selection size, it makes it hard to deselect the units in the same way, because there's often overlap when selecting more units (some of the old selected units would then be deselected). I think this would be a good opportunity to brainstorm some solutions to figure this out. Thanks for pointing it out!

arorasomya:
> Yes, we thought a bit about this. Users definitely found it easier to just click to select and click again to deselect. However, once you increase the selection size, it makes it hard to deselect the units in the same way, because there's often overlap when selecting more units (some of the old selected units would then be deselected). I think this would be a good opportunity to brainstorm some solutions to figure this out. Thanks for pointing it out!



> Yes, we thought a bit about this. Users definitely found it easier to just click to select and click again to deselect. However, once you increase the selection size, it makes it hard to deselect the units in the same way, because there's often overlap when selecting more units (some of the old selected units would then be deselected). I think this would be a good opportunity to brainstorm some solutions to figure this out. Thanks for pointing it out!

Awesome, that would be great! Approving the simplified controls for now, we can come back to make edits to the selection menu. :)

PR #445, opened 2020-10-20
Add form length limits to frontend

k-y-le:
**changes**
- frontend form textareas now only allow input up to the maximum length in the model

**to test**
- go to entry page
- try and write something over 500 characters for entry reasons areas
- try the same but 100 char for other areas etc


arorasomya (approved):
Looks good! Sorry it took so long!

PR #452, opened 2020-10-26
Address opt out + drive bugfix

k-y-le:
**changes**
- organizations can now opt out of requiring user addresses when creating / editing a drive
- a drive that doesn't have required addresses will not display the red "(required)" on the entry page, and doesn't check for addresses when validating the form
- fixed a bug where drives created in states that don't have a state page would cause an error when visiting the drive link.

**to test**
- python manage.py migrate
- python manage.py runserver
- create a drive
- deselect "require user addresses"
- test that you can submit a community without an associated address
- test same for a drive that does require user addresses - check that it fails without an address


arorasomya (approved):
Tested it out! LGTM!

Issue #448, opened 2020-10-21
No error message when logging in with incorrect credentials


k-y-le:
**Describe the bug**
Trying to log in with incorrect credentials just results in a reloaded form

**To Reproduce**
1. Go to /accounts/login
2. Try to login with incorrect credentials
3. See no error message

**Expected behavior**
An alert at the top of the form that says "incorrect username or password" or something of the sort

arorasomya:
Chuk

Chuk116:
Closing issue.

Issue #433, opened 2020-10-09
Remove zoom buttons on mobile maps


k-y-le:
**Is your feature request related to a problem? Please describe.**
Zoom buttons are not useful when one can pinch to zoom

**Describe the solution you'd like**
Zoom buttons are removed when being viewed on a phone / touch screen -- mapping apps like Google Maps don't have zoom buttons

Denisqo:
I would like to work on this role. Please assign me

arorasomya:
Chuk

Chuk116:
Closing issue.

Issue #428, opened 2020-10-09
Organizations shouldn't be able to create a drive in a state the org doesn't operate in


k-y-le:
**Is your feature request related to a problem? Please describe.**
Currently, no matter what states an organization is within, they can select any state for a community drive

**Describe the solution you'd like**
Only allow organizations to select the state(s) they operate in when creating a community mapping drive

arorasomya:
Chuk

Chuk116:
Closing issue.

PR #467, opened 2020-11-18
Login, Zoom Buttons, and Drive Creation issues fixes

Chuk116:
**changes**
- Display red bootstrap alert when a user enters a mismatched password
- Removes the zoom buttons on all of the maps when the device width is 760px or less
- Changes the state selection dropdown on the drive collection page, to only show the states that the organization is in.

**Issues Addressed**
- https://github.com/Representable/representable/issues/448#issue-726603352
- https://github.com/Representable/representable/issues/433#issue-718249746
- https://github.com/Representable/representable/issues/428#issue-718241574

**To test**
- On the login page type in a username and password that is incorrect. Should see red alert above the text boxes
- Go to a mobile phone device (< 760px) on the developer tool or use a mobile phone. Go to a map that originally had the zoom buttons. The zoom buttons should not be present.
- Create an organization and then click on the button to create a drive. In the states dropdown, you should only see the states that the organization is in.

Login-error

### DIFF
--- a/main/forms.py
+++ b/main/forms.py
@@ -178,6 +178,12 @@ class AllowlistForm(ModelForm):
 
 
 class DriveForm(ModelForm):
+    def __init__(self, org_states, *args, **kwargs):
+        super(DriveForm, self).__init__(*args, **kwargs)
+        choices = [state for state in STATES if state[0] in org_states]
+        self.fields["state"].widget = forms.Select(
+                choices=choices, attrs={"class": "form-control"}
+        )
     class Meta:
         model = Drive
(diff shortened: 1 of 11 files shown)


k-y-le (approved):
looks good!! one tiny thing to add -- removing zoom buttons from the entry page map (geo.js file) when on mobile

Chuk116:
> looks good!! one tiny thing to add -- removing zoom buttons from the entry page map (geo.js file) when on mobile

I believe that I did that already. Included in the list of edited files is the geo.js file.

PR #470, opened 2020-11-24
update small logo graphic, set default mapping radius to 0, update drive landing page

k-y-le:
**changes**
- updated lowres logo graphic to be more accurate
- set default mapping radius to 0, rather than 25/50
- updated drive landing page to match look of current state page

**to test**
- python manage.py runserver
- check logo in navbar looks as expected
- check no errors when mapping
- check drive landing page looks good

**screenshots**
image


ETedward (approved):
Looks good!

PR #478, opened 2020-11-25
Homepage map split

Chuk116:
**changes**
- Adds the entry preview page which all the "Add my community" links now point to.
- Splits off the map from the index page into its own page after the entry preview page

**to test**
- click any of the "Add my community" buttons or links and you should be redirected to the entry preview page
- click on the "Add my community" button on the entry preview page and you should be taken to the entry state selection page

**screenshots**
*Desktop*
image

*Ipad*
image

*Mobile*
image

*Entry State Selection Page*
image


k-y-le (changes requested):
two small changes 
- seems like the link from the "get started" button on the home page still goes to "#select" (line 26)
- i think the "add my community" button on the entry_preview page should be a bit bigger, imo

Chuk116:
> two small changes
> 
> * seems like the link from the "get started" button on the home page still goes to "#select" (line 26)
> * i think the "add my community" button on the entry_preview page should be a bit bigger, imo

Okay got it

k-y-le (approved):
looks good!! thx for making those changes

PR #493, opened 2020-12-04
New entry page questions

k-y-le:
**changes**
- updated entry page questions according to https://www.notion.so/Survey-Questions-0c4a8b1318114496a659e90bbc56317b

**to test**
- go to entry page
- check for typos/confusing aspects of survey questions

**of note**
- these questions being updated means that they are no longer translated! we will need to get to that for spanish relatively soon.


k-y-le:
closing this issue for now as we're still finalizing these Qs

PR #494, opened 2020-12-10
Errorhandling

arorasomya:
Changes made:
* Added error handling so if a community polygon doesn't exist in the db or aws, the page doesn't crash.
* Fixed spacing for drive and org when a community is returned from aws

To test:
* Go into postico and delete polygon for one of your entries
* go to the review page and check for a 500 error or for the community whose polygon you deleted
* also check spacing on the review page
* do the same for drive maps/org maps


k-y-le (approved):
looks good!!

PR #510, opened 2020-12-17
state name entry page bug fix

arorasomya:
Users can't access entry/* where * is a keyword that is not a state abbreviation

To test:
Go to entry/* where * is not a state abbrev and check if you are redirected to the state selection page

### DIFF
--- a/main/views/main.py
+++ b/main/views/main.py
@@ -69,6 +69,7 @@ from ..models import (
     State,
     BlockGroup,
 )
+from ..choices import STATES
 from django.views.generic.edit import FormView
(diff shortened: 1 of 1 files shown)


k-y-le (approved):
looks good!!

jasony123123:
It seems like `/entry/tx/` also redirects to `/entry_state_selection/`

arorasomya:
> It seems like `/entry/tx/` also redirects to `/entry_state_selection/`

OMG u right, the fix isn't complete! Making changes right now!

jasony123123 (approved):
looks good

PR #522, opened 2020-12-21
State Map Pages

arorasomya:
Changes:
* no access to the country-wide map page
* state map pages

image

image


k-y-le (commented):
two notes!
- on my map page, the comms are loading on the sidebar, but not on the map -- is this because of aws stuff and do i need to drawn new communities?
- we can probably go ahead and remove the "Representable is in beta" line at the top of the map page!

arorasomya:
> two notes!
> 
> * on my map page, the comms are loading on the sidebar, but not on the map -- is this because of aws stuff and do i need to drawn new communities?
> * we can probably go ahead and remove the "Representable is in beta" line at the top of the map page!

Yeah for the first note: u'd have to put polygons into the census_blocks_polygon field for them to show up on the map (they need to be in the db now and wont be fetched from aws. I can make the second note change right now!

k-y-le (approved):
ok awesome! edited some comms in the db and this appears to be working as expected

PR #528, opened 2020-12-23
make map page fetch org 0 with opt

jessiejofielding:
Changes:

- make map page use the optimized query (/map)
- use hardcoded id to get org with id 0 (on representable-dev that's vnp) so that query is for one org

To test:
- have an org with id 0 in the db and have non-aws community entries associated with that org
- run server and go to /map
- if this is a problem locally it should work really easily on staging and only affects /map and not the other maps


arorasomya (approved):
LG!

PR #535, opened 2020-12-29
Revert get calls to AWS

arorasomya:
Changes made:
* revert getting entries from aws for partner map/drive map/submission and review pages
* restored original geojson export functionality for submission/review/map pages

To test:
* create an entry in a drive
* view the map page for the drive
* view the partner map page for the org
* go to the review page and find ur entry
* export entries on the review page
* go to the submission page for the entry
* export the entry on the submission page
* go to the drive page and export entries
* go to org map page and export entries

### DIFF
--- a/main/models.py
+++ b/main/models.py
@@ -299,6 +299,7 @@ class CommunityEntry(models.Model):
     other_considerations = models.TextField(
         max_length=500, blank=True, unique=False, default=""
     )
+    # make this foreign key relation
     state = models.CharField(
         max_length=10, blank=True, unique=False, default=""
     )
--- a/main/views/main.py
+++ b/main/views/main.py
@@ -315,12 +315,23 @@ class Review(LoginRequiredMixin, TemplateView):
         approvedList = list()
         # in this case, just get the ones we made
         query = CommunityEntry.objects.filter(user=user)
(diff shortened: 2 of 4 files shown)


k-y-le (commented):
looks good to me! everything worked, except I noticed one thing

when exporting all maps as GeoJSON from the partner map page, it only exported the test community i just created, and not the other communities submitted to the same drive previously. not sure if that's a bug here or with the implementation of the multi-community export tho!

arorasomya:
> looks good to me! everything worked, except I noticed one thing
> 
> when exporting all maps as GeoJSON from the partner map page, it only exported the test community i just created, and not the other communities submitted to the same drive previously. not sure if that's a bug here or with the implementation of the multi-community export tho!

Hmm interesting. Looking into it right now!

arorasomya:
> > looks good to me! everything worked, except I noticed one thing
> > when exporting all maps as GeoJSON from the partner map page, it only exported the test community i just created, and not the other communities submitted to the same drive previously. not sure if that's a bug here or with the implementation of the multi-community export tho!
> 
> Hmm interesting. Looking into it right now!

Fixed!

k-y-le (approved):
change looks good!!!

PR #540, opened 2020-12-30
frontend forms only accept 1 state

jasony123123:
When creating a new org, only a single state can be selected. Once an org is created, the state cannot be changed anymore.

Does not change the Organization Model to force only 1 state ... since old organizations might have multiple states already attached.

To test: try creating an organization and editing an organization


k-y-le (approved):
looks good!!

PR #558, opened 2020-12-31
validation fixes

Chuk116:
- Adds a backup client-side validation in case the user alters the survey flow.

- Fixes missing recaptcha error not showing on frontend

- Fixes mapping page errors not showing up


Chuk116:
@arorasomya I just was that you have a similar pr that fixes the recaptcha. Which one do you yall think we should go with? This is how it looks on my end.

Screen Shot 2020-12-31 at 12 29 14 PM

arorasomya:
> @arorasomya I just was that you have a similar pr that fixes the recaptcha. Which one do you yall think we should go with? This is how it looks on my end.
> 
> Screen Shot 2020-12-31 at 12 29 14 PM

Hey Chuk, the recaptcha PR added functionality that validates recaptcha on the backend (preventing bots from submitting entries). I'm not sure if that is related to the validation you added? Your validation is on the front-end side, right?

arorasomya:
Also if you're worried about merging, we can hop on a call to merge tg

Chuk116:
> Also if you're worried about merging, we can hop on a call to merge tg

oh i see. Yea my own is strictly client-side changes.

PR #564, opened 2021-01-06
update wording on entry pages

k-y-le:
changed wording to entry page to match designs
before survey
image
survey questions
image
image
mapping
image

### DIFF
--- a/main/templates/main/entry_survey.html
+++ b/main/templates/main/entry_survey.html
@@ -49,7 +49,7 @@
                                 </div>
                                 <div class="col-lg-8 col-sm-11">
                                     <div class="row h-100 survey-start-txt-pic">
-                                        <div class="col-lg-12 text-left survey-start-sm-txt">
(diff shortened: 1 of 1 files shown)


Chuk116:
Maybe there could be a little bit of margin between the questions and the examples in the dropdown. Outside of that, it looks good to me!

okusio:
Agree on that. Also if we can increase the line length on the examples in
the community of interest explanation part. Also if we can switch black
text to the dark grey that would be great.

k-y-le:
Thanks for the feedback! @Chuk116 @okusio I updated with more margin + gray text; I don't think increasing the line length is too feasible without it looking strange as the text goes beyond the boxes / mismatches with the other aspects of the page... but can make it work if we think it's necessary to increase the length !

image

k-y-le:
oh @okusio did you mean on the page before the survey page? I made these changes
image

Chuk116:
Looks great!

PR #585, opened 2021-01-17
Map contiguity check for erase

k-y-le:
**changes**
- when erasing on the map, checks for contiguity
- if erase would lead to a non-contiguous selection, displays an error message and doesn't allow the user to erase

**to test**
- python manage.py runserver
- go to mapping page
- create a contiguous selection
- try to erase a block in the middle that would lead to a non-contiguous community
- check error message appears and you aren't allowed to erase

**note**
- as with the contiguity feature from before, there are occasional edge cases where you can end up with a non contiguous community. this is due to the irregular nature of census block groups

### DIFF
--- a/main/static/main/css/style.css
+++ b/main/static/main/css/style.css
@@ -144,7 +144,7 @@ form:invalid {
   background: #DFDFDF;
   width: 50px;
   height: .25rem;
-  overflow: hidden; 
+  overflow: hidden;
 }
 
 .progress-line-vert {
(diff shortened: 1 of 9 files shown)


Chuk116 (approved):
When I was testing it on the California entry page there were quite a few spots that I couldn't erase because of the check, however, based on the eye test the selection wouldn't be discontiguous. Like in the pic below when I selected the top area.
 
Screen Shot 2021-01-17 at 2 02 07 PM

I think for the most part it is fine in Cali, but it might be worth discussing maybe leaving the check till the end or changing the color of the layer rather than preventing the user from deselecting as a method of checking.

PR #587, opened 2021-01-18
fix focus console errors

k-y-le:
**changes**
- fixes console error on survey page: "An invalid form control with name='' is not focusable."
- removes some print statements

**to test**
- python manage.py runserver
- go thru entry page survey
- check error above is not visible ^^


arorasomya (approved):
Looks great!!

PR #593, opened 2021-01-19
Fixed Autoemail Merge Bug

ETedward:
So autoemailing the PDF initially worked, but then what broke it was that the code for generating the PDF for downloading changed (but the code for generating the PDF for emailing was the old one) in submissions.js.
That's fixed now! If you could check it out now — after submitting an entry, PDF should automatically be emailed!

Unfortunately there's a new bug, the PDF generated does not have all necessary polygons. I see that for the PDF downloading Kyle made some changes and added a delay before downloading to make sure the polygons are loaded ; would you be able to take a look at this / to also get the polygons loaded in time for the emailPDF() function?


k-y-le:
this is awesome!! i added a timeout function :)

i'm not sure if this is intended for debug mode or something but it appears to send the email each time I access the page -- is this supposed to be happening? I'm running on chrome, macbook 13" viewing a community I just created locally

k-y-le:
update: fixed the autoemail so that it only sends once upon submission, and updated thanks modal briefly as below

image

PR #596, opened 2021-01-22
Retroactive drive add

jasony123123:
enables a community to be added retroactively to a drive. 

A user may add a drive in the following conditions: user is logged in, viewing a community that doesn't have a drive, viewing a community they authored.

The drives that a user can add to are public, active drives in the same state as the community.

To test: 
1. create a community without a drive
2. once submitted retroactively add to a drive
3. confirm from drive dashboard that the community is added 
4. confirm that the community can not be added to anymore drives


k-y-le (commented):
looks good! i didn't encounter any errors while testing. one small change : let's change up the design of this dropdown to be a bit more in touch with the rest of the design, @okusio thoughts?

image

jasony123123:
mm ok ... how so?

okusio:
This was the original design, but ill change it up to have a drop down. Version 2 is on its way

jasony123123:
oops ignore that

PR #597, opened 2021-01-26
fix video + change contiguity check

k-y-le:
**changes**
- fixed video embed on map page
- changed contiguity check to a warning rather than a block
- small changes to formatting of video button

**to test**
- python manage.py runserver
- navigate to entry page mapping portion
- check video is visible
- check contiguity check does not block but instead warns the user

**screenshots**
image


arorasomya (approved):
My venv is not working still, but code looks good!

PR #601, opened 2021-02-11
visualizations: line thickness, colors, COI outlines

k-y-le:
**changes**
- increased the thickness of data layers on visualization pages
- removed COI outlines
- edited hover impact on COI view to be more emphatic
- added different colors for data layers (question: do we want to limit people to only having one data layer visible at a time?
- 
**to test**
- python manage.py runserver
- go to map page (preferably one with many overlapping COIs)
- check data layers, aesthetics, no errors, etc

**screenshots**
two overlapping communities with no borders:
image
hovering over one of the two:
image
congressional districts:
image
state house districts:
image

(not going to upload a pic of each layer but you get the point. each one is a different color)

### DIFF
--- a/main/static/main/js/components/keys.js
+++ b/main/static/main/js/components/keys.js
@@ -15,6 +15,17 @@ var BOUNDARIES_LAYERS = {
   "pos4": "5-digit Postcode Areas",
   "sta5": "Census Block Groups"
 }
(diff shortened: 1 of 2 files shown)


arorasomya (approved):
Can't review locally. But scanned through your code :) Looks good!

PR #605, opened 2021-02-17
Increasing accessiblity on the survey page

arorasomya:
To test:
* run python manage.py runserver
* go through the survey flow
* check whether or not the buttons are the right size and aligned

screenshots***


Chuk116 (changes requested):
The branch needs to be updated from main

PR #616, opened 2021-03-04
Anna layer branch

aceaton:
I added tribal land layers to submission and map pages. Currently using the AIANNH dataset, but may update with combination of Alaska/national datasets found at https://biamaps.doi.gov/bogs/datadownload.html. Also fixed a bug where Chicago ward/community layers weren't showing up on the submission page. Can also change the color for the tribal lands layer, right now it's pink. 
Screen Shot 2021-03-03 at 10 08 27 PM
Screen Shot 2021-03-03 at 10 09 54 PM


k-y-le:
Looks really good! I wonder if "tribal boundaries" is the best name for the data layer or if we should try to find something that makes it more clear that this is from the census? since tribal boundaries are pretty commonly disputed, i think -- curious your thoughts @aceaton !!

Also - it appears that when I submit a community to a mapping drive in illinois, the submission page doesn't show up with the chicago wards / community areas. I'm guessing that the state object isn't passed when it's a drive for whatever reason, or something like that

image

aceaton:
I added the two nyc layers and changed the name of the tribal layers to include census. I'm still working on fixing the bug with the state object being lost on drive submissions.

k-y-le:
awesome - NYC layers look good! just comment when you find the bug and i'll give it a proper review

aceaton:
All right I think the bug is gone now!

k-y-le (approved):
Looks good to me! Awesome work, Anna :)

PR #621, opened 2021-03-12
two entry page bugfixes

k-y-le:
**changes**
- fixed bug where if you pressed enter when searching for a place to start on map page, it would result in you seeing both the address portion of the survey page as well as the privacy portion at the same page (problem: buttons needed to have type="button" so they wouldn't be form focused, and added a dummy button with type="submit" in form so that pressing enter doesn't trigger form validation
- fixed bug where users couldn't select units on the same line as the "back" and "next" buttons on map page

**to test**
- python manage.py runserver
- go thru survey user flow
- on map page, search for a zip code then press enter key
- select units on map page that were on the same line as the next button
- click next
- check no errors

### DIFF
--- a/main/static/main/css/style.css
+++ b/main/static/main/css/style.css
@@ -905,10 +905,37 @@ Entry Mapping Page
 
 #map-bottom-row {
   position: absolute;
-  z-index: 1;
   bottom: 20%;
   width:100%;
   left: 0;
+}
+
+#mapToPrivacy {
+  position: absolute;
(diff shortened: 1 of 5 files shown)


Chuk116 (changes requested):
The back button on the mapping page is not working for me for all devices. Outside of that, however, things seem to be working fine.

k-y-le:
Good catch - fixed that bug

PR #622, opened 2021-03-14
Select Communities to View in "My Communities"

Chuk116:
**changes**
- On the "My communities" page, adds checkmark that gives users the ability to select particular communities to view

**to test**
- Go to the "my communities" page on an account with communities
- Verify that no communities are displayed
- Verify that upon hovering over a hidden community, it appears highlighted on the map then disappears when you stop hovering
- Verify that clicking the check box displays the community, and removing the check hides it

### DIFF
--- a/main/static/main/css/style.css
+++ b/main/static/main/css/style.css
@@ -1107,6 +1107,23 @@ Entry Mapping Page
   color: black;
 }
 
+.community-review-span:hover {
+  background-color: #dee2e6;
+  cursor: pointer;
+}
+
+.communities-menu {
+  max-height: 500px;
+  overflow: scroll;
+}
+
+@media screen and (min-width: 768px) {
+  .communities-menu {
+    max-height: 800px;
+    overflow: scroll;
+  }
+}
+
 /* landing page map */
 #usa-map {
   width: 100%;
--- a/main/static/main/js/review.js
+++ b/main/static/main/js/review.js
@@ -191,7 +191,7 @@ map.once("load", function () {
         },
       },
       layout: {
-        visibility: "none",
+        visibility: "visible",
       },
       paint: {
         "fill-color": color,
@@ -212,7 +212,7 @@ map.once("load", function () {
         },
       },
       layout: {
-        visibility: "none",
+        visibility: "visible",
         "line-join": "round",
         "line-cap": "round",
       },
--- a/main/static/main/js/review.js
+++ b/main/static/main/js/review.js
@@ -109,9 +109,22 @@ function newLowerLegislatureLayer(state) {
   });
 }
 
+function toggleEntryVisibility(checkbox){
+  if (checkbox.checked){
+    map.setLayoutProperty(checkbox.value, "visibility", "visible");
+    map.setLayoutProperty(checkbox.value + "line", "visibility", "visible");
+    hoveredOverHidden = false;
+  }
+  //If it has been unchecked.
+  else {
+    map.setLayoutProperty(checkbox.value, "visibility", "none");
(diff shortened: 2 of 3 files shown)


k-y-le (commented):
Looks good! Everything seems to be working well, nice work!

Two things that may be confusing for the average user:
- at first, the map on this page still zooms into the bounding box of the first COI on the list. But without that COI visible, it's confusing why it zooms there (Ex: mine zoomed in to central illinois right away, i didn't know why at first)
- i wonder if users might think that their community got deleted or isn't showing up if it's not visible at first? should we make it so that COIs are automatically selected and then the user deselects ones they dont want to see? not sure, what are your thoughts @Chuk116 ?

one other note on this area: @okusio i wonder if we should have our own style of checkbox? the default can tend to look a little plain and sometimes small

Chuk116:
> Looks good! Everything seems to be working well, nice work!
> 
> Two things that may be confusing for the average user:
> 
> * at first, the map on this page still zooms into the bounding box of the first COI on the list. But without that COI visible, it's confusing why it zooms there (Ex: mine zoomed in to central illinois right away, i didn't know why at first)
> * i wonder if users might think that their community got deleted or isn't showing up if it's not visible at first? should we make it so that COIs are automatically selected and then the user deselects ones they dont want to see? not sure, what are your thoughts @Chuk116 ?
> 
> one other note on this area: @okusio i wonder if we should have our own style of checkbox? the default can tend to look a little plain and sometimes small

Those are definitely good points you made. 

1) With regards to the first point, we can remove the automatic zooming I believe.
 
2) With regards to the second point, my thought process was that in the case of there being a lot of communities, it may be more annoying to deselect all the ones that you don't want to see. I think the label "Show" and the checkbox are enough indicators that they should select the checkbox to view the coi. Also, when they hover over any community, it is highlighted on the map regardless of whether it is checked or not. That being said, I am sure that user tests that Olivia runs will reveal if it is problematic. If it is, then we can change it.

k-y-le:
> Those are definitely good points you made.
> 
> 1. With regards to the first point, we can remove the automatic zooming I believe.
> 2. With regards to the second point, my thought process was that in the case of there being a lot of communities, it may be more annoying to deselect all the ones that you don't want to see. I think the label "Show" and the checkbox are enough indicators that they should select the checkbox to view the coi. Also, when they hover over any community, it is highlighted on the map regardless of whether it is checked or not. That being said, I am sure that user tests that Olivia runs will reveal if it is problematic. If it is, then we can change it.

I think I'd counter this by saying that the average user _won't_ have that many communities. Most people will only have one community drawn, if not a handful. For this page (at least) it seems to me that there would only be a small minority of people for whom it's particularly annoying to deselect communities (probably for us only, haha)

Chuk116:
> > Those are definitely good points you made.
> > 
> > 1. With regards to the first point, we can remove the automatic zooming I believe.
> > 2. With regards to the second point, my thought process was that in the case of there being a lot of communities, it may be more annoying to deselect all the ones that you don't want to see. I think the label "Show" and the checkbox are enough indicators that they should select the checkbox to view the coi. Also, when they hover over any community, it is highlighted on the map regardless of whether it is checked or not. That being said, I am sure that user tests that Olivia runs will reveal if it is problematic. If it is, then we can change it.
> 
> I think I'd counter this by saying that the average user _won't_ have that many communities. Most people will only have one community drawn, if not a handful. For this page (at least) it seems to me that there would only be a small minority of people for whom it's particularly annoying to deselect communities (probably for us only, haha)

Ah, yea that makes sense. I will switch it up then.

PR #626, opened 2021-03-16
Add verified organizations

k-y-le:
**changes**
- added field to model for an organization for verified, which defaults to false
- if verified=True, displays a checkmark next to the org's name when it appears on the site

**to test**
- python manage.py runserver
- in postico, change an organization's Verified field to True
- check for check mark where organization name appears

**screenshots**
on org dashboard
image
on state map page
image
on drive landing page
image


arorasomya (approved):
This looks great! Can we add a verified check on the org map page near its name as well?
image

Also im gonna merge this along with the more recent prs once we single out the cause of the server 500

izaller (approved):
looks so good!

PR #628, opened 2021-03-18
only allow one layer to be checked at a time

izaller:
on submission page:
https://user-images.githubusercontent.com/68291744/111668017-aaf24300-87eb-11eb-9bab-cedcd3ef634f.mov

on state map page:
https://user-images.githubusercontent.com/68291744/111668024-ae85ca00-87eb-11eb-9af4-bdbc909ce1d1.mov


k-y-le (approved):
looks great!!

PR #636, opened 2021-03-23
various bugfixes + contiguity check improvement

k-y-le:
**changes**
- fixed bug where it's not possible to erase the first block group you select
- restored contiguity check so it doesn't give warnings as much
- fixed (mostly, it's kinda finnicky because of the way mapbox does things lol) the bug where a COI could be drawn and would then be cut off -- added a delay when pressing "save" on the mapping page that pops up the "saving community..." box
- fixed missing circles on entry preview page

**to test**
- python manage.py runserver 
- draw a COI - check that you can erase the first block group you select
- select an area, drag map away from said area so that the COI is not visible, then check that COI saves appropriately

**screenshots**
circles
image

### DIFF
--- a/main/static/main/js/geo.js
+++ b/main/static/main/js/geo.js
@@ -1263,7 +1263,7 @@ map.on("style.load", function () {
         }
         else {
           selectBbox = turf.difference(selectBbox, currentBbox);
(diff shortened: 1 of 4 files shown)


arorasomya (approved):
LGTM

PR #648, opened 2021-03-24
Redid changes in clean branch

aceaton:
I turned off the prettier hook so I think that got rid of most of the problem of excess changes. I think the black hook might have still made 2 modifications to main.py but that's it. Other than that I think everything still works like it did before.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -4,22 +4,22 @@ repos:
     hooks:
       - id: black
         language_version: python3.7
-#   - repo: https://github.com/prettier/prettier
-#     rev: "2.0.5" 
-#     hooks:
-#       - id: prettier
-#         name: prettier
-#         entry: prettier --write
-#         files: "\\.(\
-#           css|less|scss\
-#           |graphql|gql\
-#           |js|jsx\
-#           |json\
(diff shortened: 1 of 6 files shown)


aceaton:
Oh wait actually, I don't really understand how precommit hooks work, but the way I turned it off was by commenting out the last two hooks in pre-commit-config.yaml, which is itself also being committed to main(?). Is that a problem? I don't want to mess up the precommit system.

.pre-commit-config.yaml
```diff
@@ -4,22 +4,22 @@ repos:
...
-          |md|markdown|mdown|mkdn\
-          |mdx\
-          |ts|tsx\
-          )$"
-  - repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v1.2.3
-    hooks:
-      - id: flake8
+#   - repo: https://github.com/prettier/prettier
+#     rev: "2.0.5"
```
arorasomya, line 8: I dont think you need to comment these out. Can you try uninstalling the pre-commit hook?

arorasomya (approved):
Left one comment, other than that, Looks great!! Thanks so much for working on this so quickly!!

PR #654, opened 2021-03-27
Select To View State Map

Chuk116:
**changes**
- Adds select to view functionality on state map page
- Adds ability to toggle back to viewing all communities at once

**to test**
- Go to state map page
- Select one or more of the checkboxes and u should only see that community(ies).
- Click the "Show All" button link and all the checkboxes should be unchecked and all the communities should be shown.


k-y-le (changes requested):
Looks good - just a couple notes:
- it seems like the "Show All" button isn't working for me. I click it and don't get an error, but nothing happens and the communities stay hidden
- it may be a good idea to make it so that if none of the communities are selected, it automatically shows all. i tested by selecting one community, then unselecting, and it was a bit confusing what to do next, since i couldn't see any communities on the map
- could we change the padding here to a margin? so that when hovering over a community testimony, it darkens the entire row
image

k-y-le (approved):
awesome - changes look great ! 

one small question: thoughts on moving the "Show All" button to the top? thinking that the more communities a state has, the more you would have to scroll in order to reach it, and most people probably wouldn't even know it exists. But then again it's not that important of a button that would be used so much, i think. Maybe @okusio has thoughts too? here's what it looks like:
image

Chuk116:
> awesome - changes look great !
> 
> one small question: thoughts on moving the "Show All" button to the top? thinking that the more communities a state has, the more you would have to scroll in order to reach it, and most people probably wouldn't even know it exists. But then again it's not that important of a button that would be used so much, i think. Maybe @okusio has thoughts too? here's what it looks like:
> image

I messed around with the placement of the "Show All" button and this was the best one I could find. I think to address the issue you were talking about I made the cois list scrollable on overflow. I will make that section smaller.

PR #655, opened 2021-03-27
need to provide admin email to add administrators to org

arorasomya:
**Changes**:
* removed the dropdown
* need to provide full email of admin
* shows error and reloads form if incorrect email or user does not exist
* redirects otherwise

**To test**:
* python manage.py runserver
* go to your org
* manage the org admins
* add more admins
* add the admin email
* check the admin page to see if they were added
* put in incorrect email and check if the error was displayed

image

### DIFF
--- a/main/forms.py
+++ b/main/forms.py
@@ -265,13 +265,16 @@ class DriveForm(ModelForm):
 class MemberForm(ModelForm):
     def __init__(self, *args, **kwargs):
         super(MemberForm, self).__init__(*args, **kwargs)
-        self.fields["member"].widget.attrs.update({"class": "form-control"})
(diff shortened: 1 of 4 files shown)


k-y-le (changes requested):
looks really good! it seems to work entirely as expected, with one exception: after logging out and logging in to another account, i saw the messages about it here
image

it's probably not a big deal to leave this, but would be confusing to a lot of orgs -- maybe we fix with an if statement so that it only shows if the admin was added?

as a side note, it would make for slightly better UX to have some sort of alert after entering an email. if i add someone, it directs me back to the dashboard, and i'm not sure if it worked or not. a little popup in the corner or something would be great!

PR #659, opened 2021-04-01
State page flow

jasony123123:
Made it look nicer as per olivia's designs.

Needs changes in Django Admin on the live site once deployed. So lmk when that happens.


arorasomya:
Would be nice to add some screenshots for context!

PR #660, opened 2021-04-02
Export Selected Cois on State Map Page

Chuk116:
**changes**
- Changes export buttons to export all the selected cois on the state map page. If no cois are selected it exports all of the cois.

**to test**
- Go to state map page.
- Try selecting specific cois then exporting them to geojson and csv. Confirm that only the cois you selected were exported
- Uncheck all the selected cois or press "Show All" then export the cois. Confirm that all the cois were exported.

### DIFF
--- a/main/static/main/js/map.js
+++ b/main/static/main/js/map.js
@@ -384,6 +384,8 @@ for (var id in toggleableLayerIds){
   var newline = document.createElement("br");
 };
 
+// Toggles the visibility of the selected community. If the coi_layer_fill layer (all the communities) is displayed, remove it and
+// display the selected community. If the last selected community community is hidden, display the coi_layer_fill layer.
 function toggleEntryVisibility(checkbox)  {
   map.setLayoutProperty('coi_layer_fill', "visibility", "none");
   if (checkbox.checked) {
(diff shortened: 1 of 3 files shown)


k-y-le (approved):
looks great! sorry for the delayed review.

some general thoughts on exporting that may be next steps (don't apply to this review, just while it's on my mind):
- it would be useful to have more context to the name of the files when exporting -- maybe go from communities.geojson -> [state]_communities_date_representable.geojson or something like that
- it would also be useful to export the list of blockgroup IDs alongside the geometry, since that would enable an easier time dealing with the occasional misaligned COI that we encountered for the law article
- we should look into making csv export align more closely with block equivalency files

Chuk116:
> looks great! sorry for the delayed review.
> 
> some general thoughts on exporting that may be next steps (don't apply to this review, just while it's on my mind):
> 
> * it would be useful to have more context to the name of the files when exporting -- maybe go from communities.geojson -> [state]_communities_date_representable.geojson or something like that
> * it would also be useful to export the list of blockgroup IDs alongside the geometry, since that would enable an easier time dealing with the occasional misaligned COI that we encountered for the law article
> * we should look into making csv export align more closely with block equivalency files

Okay those all sound like great ideas for future implementations

PR #666, opened 2021-04-14
Frontend Fixes

Chuk116:
**changes**
- Changes question button on map page from producing a dropdown to producing a modal
- Removes pictures on survey start page
- removes the footer from the survey entry flow
- fixes dropdowns on the state pages


k-y-le (approved):
looks good!

PR #667, opened 2021-04-15
add census block mapping for all states

k-y-le:
**changes**
- adds ability to switch units from block groups -> blocks for every state
- two states (illinois + oklahoma) use 2010 blocks, with population data included, all the rest use 2020 blocks and have no population
- add population button in lower right corner of mapping page (if population is available)
- adds "want different units?" button with accompanying modal explaining difference between units

**to test**
- python manage.py runserver
- create a community in any state
- click "want different units?" button on mapping page and switch to census blocks
- check you can submit a community
- check that you can submit a community comprised of block groups
- check that the population only appears in the lower right corner of the mapping page when expected (blocks + illinois/oklahoma, blockgroups + all states except nc, az, mi, nv, nj, nm, pa, ut, va, wa)

**screenshots**
illinois block group mapping
image
modal for switching
image
illinois block mapping
image


jasony123123:
no issues found

jasony123123 (approved):

k-y-le:
update: added to this PR (bc it relies on logic from adding the block mapping) the updated census block groups

after this, every state except for oklahoma and illinois will be updated for 2020 geographies

also fixed a bug with the "undo" button

Chuk116 (approved):
Everything looks good. Very small frontend nit: maybe we should take out the plus sign from the "Your Community Info" button. Maybe it's just me but it looks slightly awkward. Also when the dropdown is opened, the plus sign remains. If we want to keep it, maybe we can replace it with the fa-plus class element and have it switch to the fa-minus class element when the dropdown is opened.

izaller (approved):
no issues found

k-y-le:
> Everything looks good. Very small frontend nit: maybe we should take out the plus sign from the "Your Community Info" button. Maybe it's just me but it looks slightly awkward. Also when the dropdown is opened, the plus sign remains. If we want to keep it, maybe we can replace it with the fa-plus class element and have it switch to the fa-minus class element when the dropdown is opened.

awesome, great point - i removed the plus bc i agree that it's not really necessary here - going to merge this soon

PR #668, opened 2021-04-15
add public comment links, limit export to logged in users

k-y-le:
**changes**
- added public comment links to portals for colorado, massachusetts, missouri, montana, new jersey, oklahoma, texas, and wisconsin on submission page
- added modal when exporting pdf that links to public input portal
- limit multiexport (on state map pages and on drive map pages) to users who are authenticated -- redirects to login page when button is pressed while not authenticated

**to test**
- python manage.py runserver
- draw a community in one of the above states ^^
- submit community
- check for public comment portal link on submissino page
- export as pdf, check modal looks as expected
- log out
- go to state map page
- press export button
- check that you are redirected to login page before returning to map and able to export

**screenshots**
image
image


Chuk116:
I just realized I was not added as a reviewer for this one lol, but I did check it out. 

I noticed that when I go to the submission page of a community from the "My Communities" page or the state map page I am unable to actually access the public comments link. Can old submissions receive public comments?

jasony123123:
no bugs encountered; one question about a design decision.

For states that don't have a community input link (e.g. ND), the "Send to your state" panel with the "go to public input portal -> " text is still present. This may be confusing to users. What should we do about this?

k-y-le:
> no bugs encountered; one question about a design decision.
> 
> For states that don't have a community input link (e.g. ND), the "Send to your state" panel with the "go to public input portal -> " text is still present. This may be confusing to users. What should we do about this?

good catch, I hadn't committed the fix i made for this. It shouldn't be appearing for those states anymore.

> I noticed that when I go to the submission page of a community from the "My Communities" page or the state map page I am unable to actually access the public comments link. Can old submissions receive public comments?

old submissions should still be able to -- was your community drawn in one of the states with a portal or site open? (colorado, massachusetts, missouri, montana, new jersey, oklahoma, texas, and wisconsin). It should be working now so that if the COI is in any of these states (even previously) it should show up with a link. if not, the card won't appear on the submission page at all.

Chuk116:
> > no bugs encountered; one question about a design decision.
> > For states that don't have a community input link (e.g. ND), the "Send to your state" panel with the "go to public input portal -> " text is still present. This may be confusing to users. What should we do about this?
> 
> good catch, I hadn't committed the fix i made for this. It shouldn't be appearing for those states anymore.
> 
> > I noticed that when I go to the submission page of a community from the "My Communities" page or the state map page I am unable to actually access the public comments link. Can old submissions receive public comments?
> 
> old submissions should still be able to -- was your community drawn in one of the states with a portal or site open? (colorado, massachusetts, missouri, montana, new jersey, oklahoma, texas, and wisconsin). It should be working now so that if the COI is in any of these states (even previously) it should show up with a link. if not, the card won't appear on the submission page at all.

Okay i think the submission was in California so that may explain why the link wasn't working. However, the card did show up on the submission page.

PR #678, opened 2021-04-22
bugfix: ability to add to drive if not community author

k-y-le:
fixes bug where an authenticated user can add another user's community to a drive

**to test**
- create a community
- copy url to submission page
- log out and log into different account
- go to submission page url
- check that no "add this community to an organization" card appears

### DIFF
--- a/main/templates/main/submission.html
+++ b/main/templates/main/submission.html
@@ -163,6 +163,7 @@
             </div>
           </div>
         </div>
+        {% if is_community_author %}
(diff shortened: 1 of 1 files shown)


arorasomya (approved):
Looks great! Thanks for being so prompt with the fix!! Pushing to staging and prod

arorasomya:
One thing that came to mind, was that this should be done in the backend as well. Having validation in the backend might be a more secure fix.

PR #681, opened 2021-04-26
Re-add Google Signup/Login/Connection

Chuk116:
- Re-adds ability to login/signup using a Google account, and connect an existing account to a Google account


k-y-le (approved):
Looks good!

jasony123123:
Some thoughts for discussion.

### No Signup using Google

image

There is no sign up using Google button. Does that mean the workflow if a customer wants to use Google account is: (1) make new account using their google email, (2) navigate to settings, and (3) connect google account? If so, this is not ideal customer experience.

### Should connect be automatic?

image

If customer pressed "Login With Google" and their account is not currently linked with Google, should we automatically connect their email with Google login? Currently you get this info box. From a customer viewpoint, I think automatically linking is better. This info box creates unnessecary complexity.

### Connect with Google Appears when it is unnessecary
image

After a customer has connected their account with Google, if they navigate to the settings page, the "Connect Google" button is still present. This may present confusion to the customer: "Did it successfully connect or not?"

Chuk116:
> Some thoughts for discussion.
> 
> ### No Signup using Google
> image
> 
> There is no sign up using Google button. Does that mean the workflow if a customer wants to use Google account is: (1) make new account using their google email, (2) navigate to settings, and (3) connect google account? If so, this is not ideal customer experience.
> 
> ### Should connect be automatic?
> image
> 
> If customer pressed "Login With Google" and their account is not currently linked with Google, should we automatically connect their email with Google login? Currently you get this info box. From a customer viewpoint, I think automatically linking is better. This info box creates unnessecary complexity.
> 
> ### Connect with Google Appears when it is unnessecary
> image
> 
> After a customer has connected their account with Google, if they navigate to the settings page, the "Connect Google" button is still present. This may present confusion to the customer: "Did it successfully connect or not?"

## No Signup with Google
Regarding the first point. The login with Google automatically performs the signup process if their Google account isn't already connected. Perhaps we could rephrase the wording better.

## Should connect be automatic
The point you brought up is valid. It's something we can definitely do for the future. I think that for now, since we don't do email confirmation to register an account, it's technically possible for someone to use an email that they don't actually have the credentials for. Therefore, we would want to avoid users accidentally connecting their Google account to another person's Representable account. Maybe we should had some sort of verified button to ensure to the sure that their account has been connected. What do you think?

##  Google Appears when it is unnessecary
Perhaps there is a question of whether we want people to be able to connect multiple third-party accounts to a single Representable acocunt. But that is the reason why the connect with Google button still appears

jasony123123:
> No Signup with Google
Regarding the first point. The login with Google automatically performs the signup process if their Google account isn't already connected. Perhaps we could rephrase the wording better.

On many sites like (e.g. doordash) when you choose to sign up with Google, you don't need to create a password. So, when I thought of "signup with google" that's the feature that I thought of. I think this is a fairly important feature for customers, because it spares them the hassle of having to create & remember a new password (annoying) or re-use a password (insecure).

> Should connect be automatic
> The point you brought up is valid. It's something we can definitely do for the future. I think that for now, since we don't do email confirmation to register an account, it's technically possible for someone to use an email that they don't actually have the credentials for. Therefore, we would want to avoid users accidentally connecting their Google account to another person's Representable account. Maybe we should had some sort of verified button to ensure to the sure that their account has been connected. What do you think?

But when you choose sign-in with Google, you have to be logged into Google right? So if a user is using someone else's email and they try to sign-in with Google, they wouldn't be able to do so. 

> Google Appears when it is unnessecary
> Perhaps there is a question of whether we want people to be able to connect multiple third-party accounts to a single Representable acocunt. But that is the reason why the connect with Google button still appears

Makes sense. I'm assuming other accounts (e.g. Facebook) are on the roadmap?

Chuk116:
> > No Signup with Google
> > Regarding the first point. The login with Google automatically performs the signup process if their Google account isn't already connected. Perhaps we could rephrase the wording better.
> 
> On many sites like (e.g. doordash) when you choose to sign up with Google, you don't need to create a password. So, when I thought of "signup with google" that's the feature that I thought of. I think this is a fairly important feature for customers because it spares them the hassle of having to create & remember a new password (annoying) or re-use a password (insecure).
>
You're right that's exactly what this feature does, or is supposed to do. If a user logs in with a Google account that hasn't been registered yet, it automatically signs up the user. Perhaps some user tests will reveal whether or not people find the phrase "Login with Google" intuitive or if we should put "Signup with Google". My only issue with "Signup with Google" is that people who have already signed up with Google may think that they will be making a duplicate account if they press the button again.
> > Should connect be automatic
> > The point you brought up is valid. It's something we can definitely do for the future. I think that for now, since we don't do email confirmation to register an account, it's technically possible for someone to use an email that they don't actually have the credentials for. Therefore, we would want to avoid users accidentally connecting their Google account to another person's Representable account. Maybe we should had some sort of verified button to ensure to the sure that their account has been connected. What do you think?
> 
> But when you choose sign-in with Google, you have to be logged into Google right? So if a user is using someone else's email and they try to sign-in with Google, they wouldn't be able to do so.
> 
You're right however the issue isn't with Google but rather that I don't believe that we actually do email confirmations for each user. So technically I can use someone else's email to signup through Representable. So they try to connect with Google using that email and we automatically connect the account, then they will be connecting the account to an account that they don't own. Therefore, at least till we get some kind of email confirmation flow, it might be a good idea to have users manually connect once logged in. 

Also, the django-allauth flow is pretty self-contained so changing the flow to automatically connect may not be super trivial but tbh i haven't looked into doing it.
> > Google Appears when it is unnessecary
> > Perhaps there is a question of whether we want people to be able to connect multiple third-party accounts to a single Representable acocunt. But that is the reason why the connect with Google button still appears
> 
> Makes sense. I'm assuming other accounts (e.g. Facebook) are on the roadmap?
>
Yea I believe they potentially are

jasony123123:
> > > No Signup with Google
> > > Regarding the first point. The login with Google automatically performs the signup process if their Google account isn't already connected. Perhaps we could rephrase the wording better.
> > 
> > 
> > On many sites like (e.g. doordash) when you choose to sign up with Google, you don't need to create a password. So, when I thought of "signup with google" that's the feature that I thought of. I think this is a fairly important feature for customers because it spares them the hassle of having to create & remember a new password (annoying) or re-use a password (insecure).
> 
> You're right that's exactly what this feature does, or is supposed to do. If a user logs in with a Google account that hasn't been registered yet, it automatically signs up the user. Perhaps some user tests will reveal whether or not people find the phrase "Login with Google" intuitive or if we should put "Signup with Google". My only issue with "Signup with Google" is that people who have already signed up with Google may think that they will be making a duplicate account if they press the button again.

Why not two buttons: "Signup for Google" and "Login with Google"?

> 
> > > Should connect be automatic
> > > The point you brought up is valid. It's something we can definitely do for the future. I think that for now, since we don't do email confirmation to register an account, it's technically possible for someone to use an email that they don't actually have the credentials for. Therefore, we would want to avoid users accidentally connecting their Google account to another person's Representable account. Maybe we should had some sort of verified button to ensure to the sure that their account has been connected. What do you think?
> > 
> > 
> > But when you choose sign-in with Google, you have to be logged into Google right? So if a user is using someone else's email and they try to sign-in with Google, they wouldn't be able to do so.
> 
> You're right however the issue isn't with Google but rather that I don't believe that we actually do email confirmations for each user. So technically I can use someone else's email to signup through Representable. So they try to connect with Google using that email and we automatically connect the account, then they will be connecting the account to an account that they don't own. Therefore, at least till we get some kind of email confirmation flow, it might be a good idea to have users manually connect once logged in.
> 
> Also, the django-allauth flow is pretty self-contained so changing the flow to automatically connect may not be super trivial but tbh i haven't looked into doing it.
> 

Suppose I signup with the email chuk@gmail.com. Then, I hit sign-in with Google. Is it possible to connect to the account "chuk@gmail.com" without even logging into that Gmail account?

Chuk116:
> > > > No Signup with Google
> > > > Regarding the first point. The login with Google automatically performs the signup process if their Google account isn't already connected. Perhaps we could rephrase the wording better.
> > > 
> > > 
> > > On many sites like (e.g. doordash) when you choose to sign up with Google, you don't need to create a password. So, when I thought of "signup with google" that's the feature that I thought of. I think this is a fairly important feature for customers because it spares them the hassle of having to create & remember a new password (annoying) or re-use a password (insecure).
> > 
> > 
> > You're right that's exactly what this feature does, or is supposed to do. If a user logs in with a Google account that hasn't been registered yet, it automatically signs up the user. Perhaps some user tests will reveal whether or not people find the phrase "Login with Google" intuitive or if we should put "Signup with Google". My only issue with "Signup with Google" is that people who have already signed up with Google may think that they will be making a duplicate account if they press the button again.
> 
> Why not two buttons: "Signup for Google" and "Login with Google"?
> 
We could do that, but I believe that one button will make it look less cluttered especially since both buttons would do the same thing. Most sites that I have interacted with generally only have one of those types of buttons. What we can do potentially is change the phrase from "Login with Google" to "Continue with Google" like AirBnB's site does. What do you think
> > > > Should connect be automatic
> > > > The point you brought up is valid. It's something we can definitely do for the future. I think that for now, since we don't do email confirmation to register an account, it's technically possible for someone to use an email that they don't actually have the credentials for. Therefore, we would want to avoid users accidentally connecting their Google account to another person's Representable account. Maybe we should had some sort of verified button to ensure to the sure that their account has been connected. What do you think?
> > > 
> > > 
> > > But when you choose sign-in with Google, you have to be logged into Google right? So if a user is using someone else's email and they try to sign-in with Google, they wouldn't be able to do so.
> > 
> > 
> > You're right however the issue isn't with Google but rather that I don't believe that we actually do email confirmations for each user. So technically I can use someone else's email to signup through Representable. So they try to connect with Google using that email and we automatically connect the account, then they will be connecting the account to an account that they don't own. Therefore, at least till we get some kind of email confirmation flow, it might be a good idea to have users manually connect once logged in.
> > Also, the django-allauth flow is pretty self-contained so changing the flow to automatically connect may not be super trivial but tbh i haven't looked into doing it.
> 
> Suppose I signup with the email [chuk@gmail.com](mailto:chuk@gmail.com). Then, I hit sign-in with Google. Is it possible to connect to the account "[chuk@gmail.com](mailto:chuk@gmail.com)" without even logging into that Gmail account?

No, it isn't possible to make that connection without logging into the Gmail Account. I think that that is what we are banking on in terms of security.

jasony123123:
> We could do that, but I believe that one button will make it look less cluttered especially since both buttons would do the same thing. Most sites that I have interacted with generally only have one of those types of buttons. What we can do potentially is change the phrase from "Login with Google" to "Continue with Google" like AirBnB's site does. What do you think

This sounds good.

> No, it isn't possible to make that connection without logging into the Gmail Account. I think that that is what we are banking on in terms of security.

Then it should be fine to connect automatically. If I'm a user using my own email account, then great. If I'm a user using someone else email account, then the connection will fail as desired.
 > Also, the django-allauth flow is pretty self-contained so changing the flow to automatically connect may not be super trivial but tbh i haven't looked into doing it.

As a first iteration, I think the current setup is ok. But we should have the automatic connection in the backlog.

PR #699, opened 2021-05-05
fix retroactive drive add

k-y-le:
**changes**
- makes it so that retroactive drive add feature will only appear when expected: user is logged in, community does not belong to a drive already, user created map, and there are drives in the state

**to test**
- create community in state with drive active
- check that retroactive drive card appears
- add to drive
- check that card disappears
- create community in state with no drives active
- check that retroactive drive card doesn't appear


jasony123123:
looks good

jasony123123 (approved):

PR #722, opened 2021-05-22
State map page fix

Chuk116:
**changes**
- Adds the select to view and select to export functionalities back to the state map page

**to test**
- go to the state map page
- check and uncheck boxes and make sure that the communities are shown and hidden
- the "show all" button should display all the communities and uncheck all the boxes
- the export buttons should only export the checked communities (if any are checked) otherwise it will export all the communities


arorasomya (approved):
Works great!!! Thanks for the quick fix!!

PR #731, opened 2021-06-02
Added more characters

nshapiro23:
Changed maximum description length to 700 characters (from 500).

Actually coding very simple, had some difficulties with committing. pre-commit seemed to arbitrarily throw an error for code that I didn't touch, so I slightly changed it (don't think it should have any effect):

`drives_to_add = [d for d in all_drives if d.state==state and d.is_active==True and d.private==False]`
to      
`drives_to_add = [d for d in all_drives if d.state == state and d.is_active and not d.private]`

Testing was quite straightforward, I made 700 character-long descriptions and it worked. I also tried inspect-elementing the form and increasing the maxlength to 701 characters and then wrote a 701 character-long description.


k-y-le (approved):
looks great! ☄️

PR #735, opened 2021-06-07
copy page link on submission + other various features

k-y-le:
**changes**
- added copy link on submission page, for folks who want to share their community via the custom link of their submission
- added the WA public input portal link to submission page for COIs drawn in the state
- update wording when creating a drive for clarity on custom drive questions

**to test**
- draw a community in WA
- check public input portal link appears on submission page
- check that "copy link" button underneath export buttons works as expected

**translations note**
- this adds two things for translation: the line on the submission page about copying a link, and the line on the drive creation page which now includes an extra sentence

**screenshots**
copy link (text changes to say "Copied!" for 1.5s after clicking
image


aceaton (approved):
Looks good! I tried to fix the translations stuff but the create drive html doesn't recognize translations tags and I'm not sure how to add that doc to the translation script so I just left it for now so we can at least get these features on in English.

k-y-le:
great! merging now,

@Chuk116 @jasony123123 just as an fyi -- see anna's comment above about the translations

PR #736, opened 2021-06-07
various small feature adds

k-y-le:
**changes**
- changes wording of "use different units" button on map page to say "use smaller units" or "use larger units" depending on currently selected unit type
- TODO: figure out how to best update these values for translations, since they have to be changed in javascript
- bugfixes modal appearance when switching from blocks -> block groups
- updates about page with interns
- updates copyright to 2021
- allows users to zoom in as much as possible, while limiting searches to a max zoom to avoid confusing overzooms

**to test**
- python manage.py runserver
- draw a community
- check that above features behave as expected. change units of coi and check that units button behaves as expcted. check that searching for a location has a limited zoom while users can zoom as much as mapbox allows

**translations note**
- about page doesn't need updated translations
- button on map page needs to be translated, see todo above


nshapiro23 (approved):
Tested all the features and looks good to me! Especially the zoom feature, which I think was implemented very well

PR #739, opened 2021-06-08
fix frontend validation for survey questions

k-y-le:
**changes**
- makes it so that users cannot go to next step of survey flow if they leave all entries blank
- works for custom questions too

**to test**
- python manage.py runserver
- create a community, with no answers for survey questions
- check validation fails on frontend, stops user from advancing


nshapiro23 (approved):
looks good

PR #740, opened 2021-06-08
standardizes map pages

k-y-le:
**changes**
- updates drive and org map pages to include ability to select specific communities
- changes search bar to work for both drive and state map pages
- fixes bug where private community info showed on sidebar of state map pages
- added population (when available) to communities on state map page
- fixes bug where reported communities (drive == true and admin_approved == false) would show on the sidebar of state map pages

**to test**
- python manage.py runserver
- create a community with no drive
- create a community in same state with drive
- test making communities private and changing admin_approved in postgres db
- test both drive map page and state map page for desired behavior below
- test that behavior works as expected to select and deselect certain communities

if community is private -> shows on drive map page, but not on state map page
if community is part of a drive, but hidden (admin_approved == false) -> shows on drive map page to org admin, but not on state map page
all other communities show on both drive and state map pages


aceaton (commented):
I tested all the features and everything works on my end!

PR #741, opened 2021-06-08
Precincts layer + election data layers

aceaton:
Added precinct boundaries and election data layers to map and submission pages.

To test:
- click through layers on map pages for a state
- submit a community and click through election data layers

Precinct data is only up for some states, to test maybe use VA or TN


k-y-le (changes requested):
This looks great!! the layers are all showing as expected, and I think it adds a really great level of analysis to the site. A few small stylistic changes:
image
there's some inconsistency around the border between these three dropdowns
image
spacing between these blocks is a bit uneven
image
vertical margins between text is a little close - maybe let's add a bit more padding to the select buttons?
image
on the state pages, it can be a bit hard to see the communities under the election layers for very one-sided dem precincts. i wonder if there's anything to do here or if it's just a tradeoff we have to make because there's so much variance in opacity due to results? 
image
for those states without election data, let's add a disclaimer or something of the sort letting them know that data is not available for that state? similar to how the demographics tab has a disclaimer for communities with no demographics available until census data comes out

all small things -- the functionality itself looks great and there's so many elections to look at! I think partners are going to be really excited about this

k-y-le (approved):
stylistic changes look great!

PR #742, opened 2021-06-08
added profanity filter

nshapiro23:
added filter to screen for profanity-- must install the google client library for python which can be done by typing the following into terminal:

`pip install --upgrade google-api-python-client google-auth-httplib2 google-auth-oauthlib`

note that we will also need to install this on the representable server

Testing TODO:

- Install library `pip install --upgrade google-api-python-client google-auth-httplib2 google-auth-oauthlib`
- make map using profane language in any of the descriptions
- check the community_entry table to see if private == TRUE for that entry
- make map using non-profane language in all descriptions
- check the community_entry table to see if private == FALSE for that entry
- (optional, only do if interested) you can fiddle around with the metrics to see if you like what we consider to be profane. i spent a lot of time tweaking the settings myself, but let me know if you're interested; i can send a short Python file with that


k-y-le (changes requested):
This looks nice! It's not quite working as expected for me, but I believe it's just a small fix that I found. It appears that this code
`
            if (flagText(comm_form.data["cultural_interests"]) or 
                flagText(comm_form.data["economic_interests"])  or 
                flagText(comm_form.data["comm_activities"])  or 
                flagText(comm_form.data["other_considerations"])):                
                    comm_form.data["private"] = True`

is located within the flagText() method, when it should be outside of it.

Using print statements, I fiddled around with what does and doesn't get flagged, and it seems to work as expected! I think it's worth a walkthrough of how it works for everyone in our weekly sync, so that we all have a sense of things and can discuss any tweaks to thresholds

PR #748, opened 2021-06-09
Spanish About Page fix

Chuk116:
title


aceaton (commented):
Great! One more minor thing — I think the Spanish translations of the engineer's jobs might all just say "head of engineering." But the rest of the page looks good!

PR #756, opened 2021-06-11
Bugfixes june 11

k-y-le:
**changes**
- fixes bug where searching geocoder on map page too quickly opens help modal
- fixes bug where communities being reported by an org admin are not auto unapproved

**to test**
- python manage.py runserver
- check mapping page geocoder bug doesn't exist
- check that reporting a community as an org admin works as expected


aceaton (approved):
Looks good to me!

PR #758, opened 2021-06-14
Profanity check v2

nshapiro23:
added filter to screen for profanity-- must install the google client library for python which can be done by typing the following into terminal:

`pip install --upgrade google-api-python-client google-auth-httplib2 google-auth-oauthlib`
note that we will also need to install this on the representable server

Testing TODO:

- Install library `pip install --upgrade google-api-python-client google-auth-httplib2 google-auth-oauthlib`
- Add to environment file (type `nano ~/.bash_profile` into Terminal) the following line
`export PERSPECTIVE_API_KEY = "<key that i will send to you securely>"` save and restart Terminal
- make map using profane language in any of the descriptions, name, or title
- check if a report was made either in Postico or Django admin
- make map using non-profane language in all descriptions
- make sure a report hasn't been issued, either in Django Admin or Postico


k-y-le:
looks good! one thing i noticed: if a community does pass the profanity check, it's admin_approved field is still set to False -- is this desired behavior?

nshapiro23:
> looks good! one thing i noticed: if a community does pass the profanity check, it's admin_approved field is still set to False -- is this desired behavior?

Hmm I thought that admin_approved is true iff. an admin manually approves it.

nshapiro23:
I can obviously change it pretty easily if necessary though

k-y-le:
I think it makes sense for it to be automatically made admin_approved if it passes the profanity filter check

then if it fails, admin_approved can be set to false, and that way the community will not appear on the map pages unless we decide it should

aceaton (approved):

PR #759, opened 2021-06-15
non-approved cois don't show on state map page

k-y-le:
**changes**
- makes it so that all entries with admin_approved field set to False will not show up on the state map page
- cois with admin_approved set to false submitted to a drive will still show up (with hidden tag) to org admins on drive map pages

**to test**
- draw coi and set admin approved to false
- check state map page to see that coi doesn't appear

**note**
- this branch should only be merged following #758 and following an alteration of the prod db to make all current communities have admin_approved set to True, since the current site automatically sets this field to false for all communities drawn without following a drive link


aceaton (commented):
Screen Shot 2021-06-16 at 9 51 24 AM
Seems to hide communities correctly, seeing as the only visible one was admin approved cuz I added it to a drive. Only thing is, are the communities supposed to still be visible on the actual map itself? It hides their names but not their outlines.

k-y-le:
good catch ! i've made a change -- communities should no longer appear on the map page nor the sidebar if unapproved and either:
- on the state map page
- viewing the drive map page as a non-org-admin user

aceaton:
Looks good!! Only thing is, for some reason i couldn't retroactively add one of my COIs to a drive, even when there is a drive in that state and it's not in any yet. I made another community and this didn't happen. I think it might be something that's resolved when this gets merged with 758 too? Anyway doesn't seem to be a problem for new communities which is weird.
Screen Shot 2021-06-16 at 11 49 21 AM

PR #760, opened 2021-06-15
displays and orders communities by date

k-y-le:
**changes**
- adds a date display on state map pages in the format like "Dec. 18, 2020"
- orders communities by date on all map pages -- most recently drawn communities show at the top of the list

**to test**
- go to state map page with multiple communities drawn
- check that date appears within "more info" dropdown
- check that dates are in order of most recent first
- check the same for a drive and organization map page

**screenshots**
image


nshapiro23:
All good. Works for me

PR #761, opened 2021-06-16
Boundary labels

izaller:
at long long last...

https://user-images.githubusercontent.com/68291744/122253024-a851c900-ce80-11eb-96b6-deea3f1c8838.mov

labels for all mapbox data layers show up.
also added python script to get the relevant feature lookup data into one file.

next step will be to add labelling to school districts, specialized boundaries (ie illinois and ny specific layers)


aceaton (approved):
Looks good!! It's just meant to be implemented on the submission page, not the map page, right?

izaller:
@aceaton so i realized about 10 mins after I submitted this pr that I also need to do it on the state map page -- I can make another pr for that either on my lunch break or tn after I get off work. I don't think it's an issue though as of right now bc we don't have a flow for people to get to the map page anyways atm

PR #766, opened 2021-06-23
Refactor maps

izaller:
pulled out common code from map.js, submission.js to file map-common.js so that changes to code/addition of layers can be easily shared between the two files

all new layer additions can be done by modifying map-common.js and will be applied to both the state map and submission map pages


k-y-le (changes requested):
This is great, and sorry for the delay! It seems to be working as expected for me on the submission and state map pages, but I'm getting a couple JS errors on the drive map page:
image

nshapiro23:
This seems to be working as-intended for me as well and the two JS errors Kyle mentioned above are now gone

k-y-le (approved):
looks good!

PR #770, opened 2021-06-27
changes to export functionality

k-y-le:
**changes**
- switches export of individual communities to a zip file with the original file type (geojson or csv) and a readme (due to a bug it isn't possible at the moment to do the same for the multiexport -- working on that)
- updated storing of census block and block group manytomany field when saving a community to work properly
- updates csv file to match block equivalency format : includes BLOCKID (census FIPS code) and DISTRICT (0 for single community, counts up for multiple)
- changes the name of exported files to match more closely with the context of what's being exported: the name of the community for single export, and the name of the state/drive/org for multi export
- update municipal wards redistricting export in wisconsin to export census block ids rather than municipal wards FIPS codes

**to test**
- python manage.py collectstatic
- python manage.py runserver
- draw a community
- export community as geojson and csv, check that file looks as expected
- export multiple communities in state/drive, check that files look as expected
- draw community for municipal wards drive: http://127.0.0.1:8000/drive/tell-us-about-your/
- export community, and check that BLOCKID column in CSV has 15 characters, representing a census block
- for each of these exports, check that the name matches as expected and described above, and if necessary test with other mapping software for import capability

### DIFF
--- a/main/models.py
+++ b/main/models.py
@@ -397,6 +397,9 @@ class CommunityEntry(models.Model):
     private = models.BooleanField(default=False, null=True)
     population = models.IntegerField(blank=True, null=True, default=0)
 
+    def human_readable_name(self):
+        return self.entry_name.replace(' ', '_')
+
     def __str__(self):
         return str(self.entry_ID)
 
--- a/main/static/main/js/geo.js
+++ b/main/static/main/js/geo.js
@@ -738,6 +738,7 @@ function createCommPolygon() {
   // clean up polyFilter -- this is the array of GEOID to be stored
   polyFilter.splice(0, 1);
   polyFilter.splice(0, 1);
+  // TODO: add census blocks for dane county by matching municipal wards to blocks
   if (!dane_cty) {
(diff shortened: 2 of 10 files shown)


aceaton:
I don't think my csvs have the correct units — still appears to be block groups. I collected static and everything, and there's a readme in the file.
Screen Shot 2021-06-28 at 4 59 31 PM
Also just a question, is the block groups layer out of date with the block groups used in mapping? Bc they don't align, is the mapping 2020?
Screen Shot 2021-06-28 at 5 01 21 PM

k-y-le:
> I don't think my csvs have the correct units — still appears to be block groups. I collected static and everything, and there's a readme in the file. UPDATE - it uploads to DRA but nothing shows up on the map

That's actually behaving as expected -- I haven't added the functionality of switching from block groups -> blocks as of yet -- if you draw a community of block groups the BLOCKID values should be 15 digits (aka correct FIPS code)

> Also just a question, is the block groups layer out of date with the block groups used in mapping? Bc they don't align, is the mapping 2020?

The mapping is 2020 for all states except Illinois and Oklahoma, so that should be why there's some mismatch -- the layer we use for mapping is 2010

PR #771, opened 2021-06-27
Dashboard Redesign

Chuk116:
**changes**
- redesigns the organization dashboard
- redesigns the drive page dashboard

**to test**
- visit the organization and drive page dashboards.
- Mess with the description lengths and the private/public drive and verified/unverified organization


k-y-le (approved):
changes look great to me! only one small thing i noticed: i feel as if these buttons (the "CREATE NEW" and the three buttons on the bottom cards) could use a tad bit more padding on the bottom

Chuk116:
> changes look great to me! only one small thing i noticed: i feel as if these buttons (the "CREATE NEW" and the three buttons on the bottom cards) could use a tad bit more padding on the bottom
> image

okay yea i agree.

PR #775, opened 2021-06-29
Add survey questions as context to submission page

k-y-le:
**changes**
- adds question mark next to "Community Information" for more context
- button triggers a modal that repeats the survey questions + includes the response of the community, if there is one
- adds a few lines that will need to be translated, but the questions and the examples remain translated as they do on the survey page

**screenshots**
image
image


aceaton (approved):
nice!

PR #778, opened 2021-06-30
Cutoffs and contiguity fix

jasony123123:
To test:

1. Go to North Dakota mapping page (I only put the object for North Dakota in the S3 bucket).
2. Draw communities (based on block groups). The contiguity check should be correct now. 
3. Make sure there is no way to submit a community that ends up cutting off part of it (e.g. that weird "cutoffs" bug discovered earlier).
4. Go to other states, there shouldn't be any noticeable changes (since the S3 bucket only has ND right now)


k-y-le (approved):
looks great!

PR #784, opened 2021-07-12
fewer API calls / simpler code

nshapiro23:
testing should be pretty simple, though not sure when you can do it given problems in main. functionality should be identical to previous version, so we basically just need to test that it works the same as before. to test:

- submit profane entries (in any of the fields) and make sure that a report gets filed
- submit non-profane entries (in any of the fields) and make sure that a report doesn't get filed
- also make sure that there are no errors upon submission


aceaton (approved):
Looks good to me!

k-y-le (approved):
looks good to me as well!

PR #786, opened 2021-07-13
backups

nshapiro23:
It's been awhile, so I think it may be a good time to just PR backups. 

The main thing to test is a Python script I'll be sending to all of you, since that is what we'll be running every day to check that the hashing is consistent. I *slightly* changed the hashing method, so that's what this change to main.py is. The Python script is quite heavily commented with instructions, but let me know if you have any questions.

TODO:

- make migrations locally, run the server
- open check-hashing.py and follow the instructions on the comments there and test that the terminal prints the okay message
- edit any field of a community entry/user information on the local database and then run check-hashing.py and make sure it sends you an email saying there's an error
- make some new communities on representable and run check-hashing to make sure everything's good


aceaton (approved):
beautiful

PR #787, opened 2021-07-13
map page: search while moving the map + display # of COIs

k-y-le:
**changes**
- on map page, displays the number of communities at bottom of sidebar
- on map page, adds button "Search while moving the map" -- when toggled, only displays the currently visible communities on the sidebar (and updates community number displayed)

**to test**
- python manage.py runserver
- go to state map page
- scroll around and see that behavior is normal
- click "Search while moving the map" button
- see that only visible communities show on sidebar, and that number of COIs is updated
- de-select button
- see that all communities re-appear on sidebar, and that dragging the map does not alter what's visible or the number at the bottom of the list
- test all the same for a drive/org map page

**screenshots**
comm number: 
image
search button (inactive):
image
search button (active):
image


nshapiro23:
looks good to me, though i won't approve since my maps is kind of buggy

PR #788, opened 2021-07-19
Custom drive

aceaton:
currently there are some css issues (with which help would be appreciated) but other than that i think it should work??


aceaton:
oh also translating thing isn't there yet bc its being bad

aceaton:
To review
first of all this is where we discussed all the changes: 
https://www.notion.so/princeton/Org-Customizability-6da6c85d6c394c58b1e07609447d81d8

makemigrations, migrate, runserver
make orgs
- both government and non-government
- with/without images 
make drives
- for government orgs the form should have the option to edit redistricting info
- for non-government it shouldn't
- use different units
make communities in those drives
- you should see logo if its there
   - it should link to org page
- should default to the units you selected
- see custom info on the first page and also COI definition later on in the survey flow
   - if you didn't customize info or logo they should just be normal

let me know if any of the functionality works weirdly or anything looks ugly

k-y-le:
some small notes: 

1. the alignment of this checkbox is slightly off (weirdly, on the third time creating a drive, it was no longer off, not sure what's going on there, but probably something we can change later)
image
2. this is more a design thing (@okusio ) but i feel that this example text could be a bit more verbose to encourage descriptive descriptions 
image
3. I'd like to see slightly more padding between the title of the form fields and the fields themselves -- also, should the field be "Mapping Units" or "Default Mapping Units" ?
image
4. I think the slogan text ("Gerrymandering is silencing communities") on the drive landing page should be moved down a bit more 
image
5. not a big deal but worth noting that the form fields when editing a drive don't have consistent appearance with the form fields when creating a drive 
image
6. uploading images with different dimensions can cause a bit of ugly formatting on the drive page (i don't think this is an immediate concern)
image
image
7. this aspect of editing an organization's details isn't super clear what's going on w.r.t. changing the drive logo 
image

these are all minor tho, the actual functionality works great for me! so going to approve

k-y-le (approved):
looks good !

PR #792, opened 2021-07-21
remove unused images + add link cover

k-y-le:
**changes**
- removed unused (old) images from codebase
- added a link cover image for link previews on social media

**to test**
- python manage.py collectstatic
- python manage.py runserver
- check that no images are missing


aceaton (approved):
I looked through every page and they all looked identical! What's a link cover though?

PR #793, opened 2021-07-21
tags update - modal with all tags

k-y-le:
**changes**
- added a dozen new tags to addtags command
- added ability to view all tags by pressing a "+" button next to the top 15 tags
- pressing said "+" button opens a modal with all tags and a search bar to find new tags
- added visual cues (active button) to tag buttons when selected or deselected
- all tags not selected will be grayed out on modal and on top15 once five tags have been added

**to test**
- python manage.py addtags
- python manage.py runserver
- go through survey flow, add tags, play around with feature to see that it works as expected

**screenshots**
image
image
image


aceaton (approved):
Great tags! 😍

PR #796, opened 2021-07-23
Switch faq + glossary to static html

k-y-le:
**changes**
- makes FAQ and glossary into static html and removes models / django admin

**to test**
- python manage.py migrate
- python manage.py runserver
- check faq and glossary pages


aceaton (approved):
Looks good!!

nshapiro23:
Also looks good to me!

PR #798, opened 2021-07-26
state resources pages

k-y-le:
**changes**
- adds a dropdown to resources page for state selection
- adds urls for state specific resources page -- on these pages, see state specific redistricting criteria, coi definitions, and process for all 50 states

**to test**
- python manage.py runserver
- go to resources page
- use dropdown
- check behavior is as expected

**screenshots**
regular resources page:
image
with state info
image


aceaton:
Nice! The only thing is it's kind of unclear to me that the line is something I can click on to select a state if I didn't already know that, like could there be grayed out text that says "choose state" or something?

aceaton (approved):
Looks good, just see my one comment.

PR #799, opened 2021-07-27
aws s3 bucket for media

aceaton:
aws buckets for logo pics
to test:
- pip install -r requirements.txt
- make an org with a pic
- make a drive
- see if that drive shows the picture that's uploaded

### DIFF
--- a/representable/custom_storage.py
+++ b/representable/custom_storage.py
@@ -1,0 +1,6 @@
+from storages.backends.s3boto3 import S3Boto3Storage
+
+class MediaStorage(S3Boto3Storage):
+    bucket_name = 'representable-media'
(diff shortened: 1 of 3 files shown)


k-y-le (approved):
looks good to me 

just a couple notes that i'm putting here to make sure it's out there - no rush though
1. i think the required asterisks when creating a drive/org should be red
image
2. I'd like to see somewhere to explain and recommend census block groups vs blocks when orgs are selecting their default mapping unit. Maybe a small question mark with info when you hover/click on it?
3. the org creation form says "states" but you can only select one -- however in our models states is an array, so orgs should be able to select multiple states
image

PR #800, opened 2021-07-27
couple minor style things

aceaton:
- added arrow to why do we need your info
- made state portal link open in a new tab


k-y-le (approved):
🏋️

PR #804, opened 2021-07-30
map page links on home page and submission page

k-y-le:
**changes**
- adds a link to the map pages on the home page
- changes the urls for state selection, so that going to /state_selection/map leads to the map pages
- adds a link to the map pages on the submission pag
- changes the urls for map pages, so that if it includes a valid url in the format /map/<state>/<lat>/<lng> then it zooms into that spot
- therefore, clicking on the map page link on the submission page will automatically zoom in to show communities that are in the area that the original community was drawn in

**to test**
- python manage.py runserver
- navigate to a comminity submission page
- check that link to map page works as expected
- check that url for map page redirects to state map page url if invalid values for the latitude and longitude
- navigate to home page
- click on "view communities" and see that it redirects to state selection page
- check that state selection page for map pages and for entry page are both working as expected
- some small cosmetic changes for mobile compatibility

**screenshots**
on submission page:
image
home page:
image
home page (mobile):
image


aceaton (approved):
peachy keen!

PR #805, opened 2021-07-30
Mixpanel added tracking

aceaton:
I added a bunch of tracking stuff. I'm not sure how to test other than just to go through making a community and make sure stuff isn't messed up? I don't think this would impact anything visibly and I can't test properly until it's on staging unfortunately.


k-y-le (approved):
looks good! after pushing to prod we can go through the site to test if it's all working as expected

PR #808, opened 2021-08-05
Added upload map feature

aceaton:
adds ability to upload maps as data layers
also fixes election data on org and drive map pages
to test:
check that custom map works on map page, submission page, org and drive maps, and that the other data layers appear as they should.


k-y-le (approved):
🐒

PR #815, opened 2021-08-16
Visualize multipolygons (2 rings)

k-y-le:
**changes**
- fixes a bug (sorta) where COIs with multiple loops

**to test**
- draw a COI with multiple pieces (easiest way to find this is to poke around the Dane County wards -- http://127.0.0.1:8000/drive/tell-us-about-your/ )
- check that two pieces of the COI are visible on the submission page
- check state and drive map pages for accurate visualizations

**screenshots**
image

### DIFF
--- a/main/static/main/js/map.js
+++ b/main/static/main/js/map.js
@@ -122,8 +122,15 @@ map.on("load", function () {
   for (coi_id in coidata) {
     // set the coordinates of the outer ring to final
     final = [];
+    featureType = "Polygon";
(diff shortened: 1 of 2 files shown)


aceaton (commented):
Screen Shot 2021-08-20 at 12 43 04 PM
looks good on all the pages, only thing is it does only show the largest two chunks, when in the case of this community at least it has like 10, but that might be the intended functionality
Screen Shot 2021-08-20 at 12 45 10 PM

PR #818, opened 2021-08-23
New census files

aceaton:
Handles new census data files
- changes feature referenced for block/bg id
- allows population display on any state with new data

test
- go to new mexico and map with both blocks and block groups, try switching, etc
- make sure it submits and everything, and population works as expected

okay SO i made mbtiles for all the states, but only uploaded nm so far for testing because there's no one there so hopefully it won't mess anything up, anyway the issue is that the files actually have weird geoids now so if i uploaded them and didn't change prod it would break (like the ok/il thing). So either they have to be uploaded concurrently with changes to prod, or i have to except for the length of the geoid every time it's referenced, which i started doing but was afraid it would be buggy and overcomplicated because it happens in so many places (though let me know if you can think of a better way to do this because I'm probably missing something dumb). I could do ~10-state batch uploads over a couple nights or something. We could also upload the files under different names than the old ones. Anyway they're all ready so just lmk what to do.

### DIFF
--- a/main/static/main/js/components/keys.js
+++ b/main/static/main/js/components/keys.js
@@ -67,7 +67,7 @@ var NYC_STATE_ASSEMBLY_KEY = "61wjirls";
 var PRECINCTS_KEY = "precincts_final";
 
 var STATES_USING_OLD_UNITS = ["il", "ok"];
-var STATES_WITH_NEW_FILES = ["nm"];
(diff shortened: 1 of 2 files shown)


k-y-le:
I think it makes sense to name these files something different - that way we can get them all out in one push to prod, without having to do too many exceptions

I'm going to take a look at this soon and will report back with any comments/feedback!

PR #872, opened 2021-10-26
integrated custom layer while mapping into drive db entry

aceaton:
to test
- go to admin and change the draw_layer parameter of any drive to a map box key and then that should show up when drawing the map
- for commerce city drive you can use the key ccbounds
when i merge to prod i have to change the drive property for that one


k-y-le (approved):
looks good!

PR #883, opened 2021-12-10
zoom to feature for drives, can add in admin

aceaton:
check by changing zoom x and y in the django admin for a drive, then the drive should start at that location


k-y-le:
causes a bug (map doesn't load) when I enter in the coordinates as [Lat, Long], instead of [Long, Lat] -- ex: [41.8, -87.6] vs. [-87.6, 41.8]

k-y-le (approved):
looks good! with just a bit more clarity on lat vs. long for coordinates, so we don't accidentally end up in antarctica again

aceaton:
Screen Shot 2021-12-13 at 3 10 10 AM

PR #922, opened 2025-09-16
Docker-compose, allauth.google as temp fix, and geo.js

zhe-CUNY:
- Adds docker files for local development.
- Updates some pip packages and removes unneeded ones 
- Use allauth.google as a sign up as a temp fix to sendgrid
- Fixes outdated turf.js (union, difference) that causes selections/ eraser not to work on the map


zhe-CUNY:
image

k-y-le (commented):
Looks good to me!

zhe-CUNY:
Thanks. I don't seem to have the permissions to merge :) So go ahead.

zhe-CUNY:
It is working https://www.representable.org/accounts/signup/

However, there is a small bug (probably a setting needs to be changed) where if you already have an existing django account and you use google to login.. it doesn't work

zhe-CUNY:
Fixed-ish. This page seems to be un-themed. Might be worth updating all-auth to a later version.

image

